PR #169, opened 2021-05-11
Spec block header

### LINKED ISSUE #145: Consider Merkle Mountain Range for Previous Block Hash
## Motivation

While the Block Header format is still under discussion, we should consider a Merkle Mountain Range as the previous block header, this way we can show more efficiently the children of previous blocks and avoid potential forking issues with referenced blocks during fraud proving.

adlerjohn:
Add format for block header, including accumulator for past block header hashes. Fixes #145.

### DIFF
--- a/specs/protocol/block_header.md
+++ b/specs/protocol/block_header.md
@@ -1,0 +1,13 @@
+# Block Header
+
+This document specifies the on-chain block header format.
+
+## BlockHeader
+
+| name        | type      | description                                                                       |
+|-------------|-----------|-----------------------------------------------------------------------------------|
+| `height`    | `uint32`  | Fuel block height.                                                                |
+| `number`    | `uint32`  | Ethereum block number.                                                            |
+| `prevHash`  | `bytes32` | Block header hash of the previous block.                                          |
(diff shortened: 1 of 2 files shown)


specs/protocol/block_header.md
```diff
@@ -0,0 +1,13 @@
+# Block Header
+
+This document specifies the on-chain block header format.
+
+## BlockHeader
+
+| name        | type      | description                                                                       |
+|-------------|-----------|-----------------------------------------------------------------------------------|
```
Voxelot, line 8: There should also be a coinbase field right?

specs/protocol/block_header.md
```diff
@@ -0,0 +1,13 @@
...
+| `height`    | `uint32`  | Fuel block height.                                                                |
+| `number`    | `uint32`  | Ethereum block number.                                                            |
```
Voxelot, line 10: Should we also specify the ethereum block hash to avoid the risk of a fork?

adlerjohn: Depends. L1 block hash isn't ideal, because Ethereum only lets you see the previous 256. It might be better to instead do something like some tip of deposits/valset updates commitment?

adlerjohn:
Closing as stale

PR #257, opened 2021-12-31
Name readme title after Ethereum yellow paper

adlerjohn:
For the lulz

Yellow paper for reference: https://ethereum.github.io/yellowpaper/paper.pdf


Voxelot (approved):
LGTM

PR #259, opened 2021-12-31
Color -> asset ID.

### LINKED ISSUE #214: Rename "color_of_coins" for more clarity
The term  "color" with respect to coins has a history in the bitcoin world, but it's not immediately clear what it means.
Simply using `token_id` (or `token_type`) might be slightly more effective in terms of onboarding new developers.
Note that this would also involve renaming `const ETH_COLOR`, possibly to something like `ETH_ID` (or `ETH_TYPE`).

This would result in the following:
`if color_of_coins == ETH_COLOR {...`
becoming `if token_id == ETH_ID {...`

or similarily:
`if token_type == ETH {...`

An additional consideration is NFTs. If native support for non-fungibles is not planned for Fuel, then using `token_id` is probably fine. Otherwise, it could be confusing in the context of an NFT which also has a `token_id`.

It's a small detail, but the sum of many such details could be the difference in the language feeling easy to pick up for devs coming from the Eth/Solidity world.

adlerjohn:
Fixes #214

This will probably require change in the ABI consumption (@digorithm), ABI generation (@emilyaherbert), general naming in the node (@vlopes11 @Voxelot), and the standard library (@nfurfaro @sezna).

### DIFF
--- a/specs/chains/ethereum/compressed_tx_format.md
+++ b/specs/chains/ethereum/compressed_tx_format.md
@@ -112,11 +112,11 @@ This document specifies the _compressed_ transaction format, which is posted to 
 
 ### OutputCoin
 
-| name           | type                                            | description                                       |
-|----------------|-------------------------------------------------|---------------------------------------------------|
-| `toPointer`    | [DigestRegistryPointer](#digestregistrypointer) | Receiving address pointer or script hash pointer. |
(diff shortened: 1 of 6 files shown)


nfurfaro (dismissed):
LG.

Voxelot (approved):

sezna (approved):

Voxelot (approved):

PR #263, opened 2022-01-07
add byte pricing independent of gas price

### LINKED ISSUE #262: Add byte price field
There is currently no mechanism to account for the bytes a tx consumes. Add a byte price field, similar to gas price, whose product with the transaction size in bytes is subtracted from the free balance.

Note that this will probably require some form of byte limit for blocks in addition to a gas limit.

Note that the introduction of a byte price removes the need for a gas price in contract creation txs.

Voxelot:
closes #262

Add a separate bytePrice field to transaction specs, which allows script execution to be priced separately of bytes.

Open question: Should gas be renamed to something more specific like scriptGas or scriptPrice?

### DIFF
--- a/specs/protocol/tx_format.md
+++ b/specs/protocol/tx_format.md
@@ -89,9 +89,9 @@ enum  ReceiptType : uint8 {
 
 | name               | type                    | description                              |
 |--------------------|-------------------------|------------------------------------------|
-| `bytePrice`        | `uint64`                | Price per transaction byte.              |
 | `gasPrice`         | `uint64`                | Gas price for transaction.               |
 | `gasLimit`         | `uint64`                | Gas limit for transaction.               |
+| `bytePrice`        | `uint64`                | Price per transaction byte.              |
 | `maturity`         | `uint32`                | Block until which tx cannot be included. |
 | `scriptLength`     | `uint16`                | Script length, in instructions.          |
 | `scriptDataLength` | `uint16`                | Length of script input data, in bytes.   |
--- a/specs/protocol/tx_format.md
+++ b/specs/protocol/tx_format.md
@@ -24,7 +24,6 @@
 
 | name                        | type     | value | description                                   |
 |-----------------------------|----------|-------|-----------------------------------------------|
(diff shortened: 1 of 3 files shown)


specs/protocol/tx_format.md
```diff
@@ -90,6 +89,7 @@ enum  ReceiptType : uint8 {
 
 | name               | type                    | description                              |
 |--------------------|-------------------------|------------------------------------------|
+| `bytePrice`        | `uint64`                | Price per transaction byte.              |
```
adlerjohn, line 92: Put this in the same order as the compressed tx format.

specs/protocol/tx_validity.md
```diff
@@ -106,7 +106,7 @@ def unavailable_balance(tx, col) -> int:
     """
     sentBalance = sum_outputs(tx, col)
     gasBalance = gasPrice * gasLimit
-    bytesBalance = size(tx) * GAS_PER_BYTE * gasPrice
+    bytesBalance = size(tx) * bytePrice
```
adlerjohn, line 109: Can you

1. remove the `GAS_PER_BYTE` entry in the constants table in `consensus.md`
2. clarify in a comment that `size()` does not include witness data (otherwise the block proposer could just include a bunch of witness data and make users pay for it)
3. fix the correct change validity rules to account for cost of bytes

?

adlerjohn (changes requested):

adlerjohn:
> Open question: Should gas be renamed to something more specific like scriptGas or scriptPrice?

Hmmm. We can bikeshed! I feel like coming from Ethereum `gas` alone is sufficient, but I'm not married to the name.

specs/protocol/tx_validity.md
Voxelot: Addressed 1 & 2. For 3 I didn't explicitly mention the bytes since their cost is already implicitly included in the initial free balance.

adlerjohn (approved):
This will need to be implemented in `fuel-tx` and `fuel-core`. It will also affect some constants in the compiler/stdlib cc @sezna @nfurfaro

Issue #201, opened 2021-08-26
Domain separate predicate hash, address and blocks


adlerjohn:
Currently, the predicate hash and address (pubkey hash) aren't domain separated. This means if a pubkey happens to have a predicate bytecode that can evaluate to `true` with some easy-to-find parameters, then anyone can spend the UTXO. Getting the predicate hash and the address should therefore be domain-separated.

Voxelot:
Now that UtxoId's are based on transaction and output index, the above domain separated hash scheme may need some changes.

MitchTurner:
Is this something we want before mainnet? @xgreenx @Voxelot 
If not, we can remove from Fuel Project Board.

Voxelot:
Correct @xgreenx, this particular issue was solved when I started down the path of including [chain id's](https://github.com/FuelLabs/fuel-specs/pull/462) into the predicate id. Even though we decided not to use chain ids there anymore, it still uses domain separation based on the contract id domain separator:
> `0x4655454C`

PR #266, opened 2022-01-09
Add rule to verify predicate hash

### LINKED ISSUE #264: Fully specify p2sh verification of predicates
The specs reference p2sh when discussing predicate verification, but the exact requirements for this aren't currently clear. We should update the predicate validation specs to include a section about how the hash of the predicate script must match the input owner.

adlerjohn:
Fixes #264

For now, compute the predicate root computed the same way as the contract ID. Eventually it will be domain separated differently with #201.


Dentosal (commented):
LGTM, but deferring to @vlopes11

Voxelot (approved):

PR #275, opened 2022-01-29
Convert txo pointer to tx pointer.

### LINKED ISSUE #273: Remove redundant outputIndex from TXOPointer
Since #253, `uint8 outputIndex` is specified directly in the [Input](https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/tx_format.md#inputcoin), so is redundant with the same information in the [TXOPointer](https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/tx_format.md#txopointer)

The `TXOPointer` type should now be just `{blockHeight, txIndex}`.

adlerjohn:
Fixes #273

Implementation:
- [ ] Client (https://github.com/FuelLabs/fuel-tx/issues/73)
- [ ] Contracts (https://github.com/FuelLabs/fuel-v2-contracts/issues/74)

### DIFF
--- a/specs/protocol/tx_format.md
+++ b/specs/protocol/tx_format.md
@@ -18,7 +18,7 @@
   - [OutputVariable](#outputvariable)
   - [OutputContractCreated](#outputcontractcreated)
 - [Witness](#witness)
-- [TXOPointer](#txopointer)
+- [TXPointer](#txpointer)
 
 ## Constants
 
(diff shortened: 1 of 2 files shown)


simonr0204 (dismissed):

Voxelot:
Should we add this field to the [post-condition validity ](https://github.com/FuelLabs/fuel-specs/blob/1c87e3a223d79f765029f31a46257507115ed042/specs/protocol/tx_validity.md#vm-postcondition-validity-rules) section? I'm assuming it's set by the block producer?

adlerjohn:
Hmmm. Not the post-condition, but the pre-condition (the UTXO should exist, and it should be pointed to correctly). There's already a line there, but it wasn't updated:

`If this check passes, the UTXO ID (txID, outputIndex) fields of each contract input is set to the UTXO ID of the respective contract. The txoPointer of each input is also set to the TXO pointer of the UTXO with ID utxoID.`

adlerjohn:
Asking @Voxelot for re-review

Voxelot (approved):

PR #281, opened 2022-02-20
bytePrice validation

rakita:
@adlerjohn i think this is how we agreed on?

### DIFF
--- a/specs/protocol/tx_format.md
+++ b/specs/protocol/tx_format.md
@@ -24,6 +24,7 @@
 
 | name                        | type     | value | description                                   |
 |-----------------------------|----------|-------|-----------------------------------------------|
+| `GAS_PER_BYTE`              | `uint64` |       | Gas charged per byte of the transaction.      |
 | `MAX_GAS_PER_TX`            | `uint64` |       | Maximum gas per transaction.                  |
 | `MAX_INPUTS`                | `uint64` | `8`   | Maximum number of inputs.                     |
 | `MAX_OUTPUTS`               | `uint64` | `8`   | Maximum number of outputs.                    |
@@ -110,6 +111,7 @@ Transaction is invalid if:
 - Any output is of type `OutputType.ContractCreated`
 - `scriptLength > MAX_SCRIPT_LENGTH`
 - `scriptDataLength > MAX_SCRIPT_DATA_LENGTH`
+- `bytePrice != gasPrice * GAS_PER_BYTE`
 
 Note: when signing a transaction, `receiptsRoot` is set to zero.
 
@@ -156,6 +158,7 @@ Transaction is invalid if:
 - Any contract with ID in `staticContracts` is not in the state
 - The computed contract ID (see below) is not equal to the `contractID` of the one `OutputType.ContractCreated` output
 - The [Sparse Merkle tree](./cryptographic_primitives.md#sparse-merkle-tree) root of `storageSlots` is not equal to the `stateRoot` of the one `OutputType.ContractCreated` output
+- `bytePrice != gasPrice * GAS_PER_BYTE`
 
 Creates a contract with contract ID as computed [here](./identifiers.md#contract-id).
 


specs/protocol/tx_format.md
```diff
@@ -110,6 +110,7 @@ Transaction is invalid if:
 - Any output is of type `OutputType.ContractCreated`
 - `scriptLength > MAX_SCRIPT_LENGTH`
 - `scriptDataLength > MAX_SCRIPT_DATA_LENGTH`
+- `bytePrice != 0`
```
Voxelot, line 113: I think we had talked about adding a consensus rule s.t. bytePrice == gasPrice * N. I don't think we should prevent fee-less txs if miners are willing to include them.

specs/protocol/tx_validity.md
```diff
@@ -106,12 +106,10 @@ def unavailable_balance(tx, col) -> int:
     """
     sentBalance = sum_outputs(tx, col)
     gasBalance = gasPrice * gasLimit
-    # Size excludes witness data as it is malleable (even by third parties!)
-    bytesBalance = size(tx) * bytePrice
     # Only native coin can be used to pay for gas
     if col != 0:
         return sentBalance
-    return sentBalance + gasBalance + bytesBalance
+    return sentBalance + gasBalance
```
Voxelot, line 112: Similarly to my last post, I'd think we'd keep the byte balance costs included but just constrain to some fixed ratio of the gas price. Otherwise, we'd only meter txs based on execution and not capture the data cost anywhere.

specs/protocol/tx_format.md
rakita: We talked about both, but it is slightly better in my oppinion to have zeroed bytePrice as it can be joker field in future if needed. But it does not matter a lot both are consensus rule change.

Voxelot: Ah I see, I misread this as requiring bytePrice to be non-zero instead of always setting it to zero.

specs/protocol/tx_validity.md
rakita: A little bit different from above, tx size usually falls under `gasLimit*gasPrice` as it is calculated as a gas it allows us to tweak it where gas is calculated. but I can revert that here and make a special rule for size as it was before.

specs/protocol/tx_format.md
rakita: Yea that rule logic is kinda weird

specs/protocol/tx_validity.md
Voxelot: Before the addition of the bytePrice field, there was a constant in the specs for GAS_PER_BYTE and it was calculated here.

Voxelot:
I vote for just adding a consensus rule to the PoS network that bytePrice = gasPrice * GAS_PER_BYTE, and leaving the specs as is. That way we don't have to make drastic changes to the VM and node for things like change amounts over and over.

specs/protocol/tx_format.md
adlerjohn: I'm okay with either here, though I'll defer to @Voxelot who says we should use [`bytePrice = gasPrice * GAS_PER_BYTE`](https://github.com/FuelLabs/fuel-specs/pull/281#issuecomment-1048331577).

specs/protocol/tx_validity.md
adlerjohn: Yeah we should keep the calculation here, just set rules on `bytesBalance` implicitly by containing `bytePrice`.

adlerjohn:
The reason I vote for `bytePrice = gasPrice * GAS_PER_BYTE` rather than `bytePrice = 0` is because if we set it to zero then we may not test the client with different prices and things might break suddenly. At least if we set it to some constant that will usually be non-zero IRL, then the codepath to compute `bytesBalance` gets actually tested.

rakita:
sure, lets go with that.

Voxelot (approved):

adlerjohn (approved):

PR #283, opened 2022-02-26
Update Array ABI spec

QuinnLee:
Updating Array to be `[T; n]` instead of `T[n]`  to match the Sway output

### DIFF
--- a/specs/protocol/abi.md
+++ b/specs/protocol/abi.md
@@ -63,7 +63,7 @@ Here an example containing custom types:
                             },
                             {
                                 "name":"b",
-                                "type": "u8[2]"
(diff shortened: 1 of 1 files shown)


adlerjohn:
I don't inherently mind either alternative, but can we confirm the Sway compiler emits `[T; n]` and that it won't change? cc @emilyaherbert

digorithm:
> I don't inherently mind either alternative, but can we confirm the Sway compiler emits `[T; n]` and that it won't change? cc @emilyaherbert

There was a conversation recently in one of our channels where @otrho confirmed that `[T; n]` is the way for types other than strings.

PR #288, opened 2022-03-03
Always append a result receipt on revert.

adlerjohn:
instead of only when in an external context


Voxelot:
Currently the VM automatically appends the ScriptResult receipt at the end of executing any kind of script, it's not handled within the opcode impl itself.

Voxelot (approved):

PR #290, opened 2022-03-07
Clarify registers used in CALL.

### LINKED ISSUE #282: Request to clarify the meaning of the registers used by the CALL opcode
For all other opcodes, it was trivial for me to learn what each register means. For the CALL opcode, it wasn't as trivial: 
* For `$rA`, we say "Register `$rA` is a memory address from which..." which is clear.
* For `$rB`, we say "increase balance of asset ID `MEM[$rC, 32]` by `$rB`" which is clear but a bit indirect.
* For `$rC`, we hint to the fact that `$rC` is basically a memory address that points to the asset ID.
* For `$rD`, we say "`$rD` is the amount of gas to forward." which is clear. 
It would be ncie to either describe the "Operation" in the table in terms of these registers (like we do for other opcodes) as much as possible, or at least list the meaning of each register right after table.

adlerjohn:
Fixes #282

### DIFF
--- a/specs/vm/opcodes.md
+++ b/specs/vm/opcodes.md
@@ -1071,7 +1071,7 @@ Register `$rA` is a memory address from which the following fields are set (word
 | 8     | `byte[8]`  | `param1` | First parameter.     |
 | 8     | `byte[8]`  | `param2` | Second parameter.    |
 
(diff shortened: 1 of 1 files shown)


mohammadfawaz (approved):
LGTM! Thank you for doing this.

Issue #289, opened 2022-03-04
ScriptResult should reflect status of RVRT in addition to panics


Voxelot:
Currently, the ScriptResult receipt only reflects errors caught by the VM (panic), but if a contract developer uses RVRT to indicate some failure state, the ScriptResult considers this as a success.

While users should look for the revert or panic receipt, I think it would be a better UX if the script result also reflected the status of user-defined errors (RVRT) in addition to panic receipts.

nfurfaro:
I feel like a revert should be presented to the user/dev as a failure, not a success.
The only exception (and not really) would be a test that is expected to revert with a specific message or error code.

Voxelot:
In the case of a negative test, it would need to look for the revert receipt to see if a specific revert code was given. ScriptResult is currently specified in the VM as more of a boolean success/failure and isn't intended to propagate the actual panic reasons or error codes (yet?).

ControlCplusControlV:
Smart Contracts should do one thing well and revert otherwise. I think it makes sense for revert to be allowed to be considered a success state as with things like negative tests for something like a test to ensure permissions are properly revoked, etc

Voxelot:
Typically when writing a negative test, you're expecting a specific failure mode. For example in Rust, if you wanted to ensure some kind of validation code worked you'd verify the error returned using something like:
```
#[test]
fn test {
    let result = some_operation(bad_input);
    assert!(matches!(result, Err(BadInputError)));
}
```
It seems strange to be writing negative tests where the transaction result is successful. While not exactly equivalent since we're talking about receipts, but in Rust you wouldn't wrap an error response inside of an `Ok` variant to make it easier to write negative tests.

ControlCplusControlV:
Was confused in regards to where success is interpreted, reread though and definitely agree it makes more sense to consider it a failure and return the RVRT error message if anything was included so the test can handle it as needed

PR #293, opened 2022-03-25
ScriptResult Error on Revert

Voxelot:
ScriptResult should reflect that revert is also an error state, instead of only panics.

closes: #289


adlerjohn:
Could you also fix the above paragraph to append the receipt regardless of context?

adlerjohn (approved):

PR #300, opened 2022-04-03
fix: use `uint32` type for all `maturity` fields

vnepveu:
- One field used `uint64` type for `maturity` field, whereas two others
  were using `uint32`.


adlerjohn (approved):

adlerjohn:
We should probably have a `BlockHeight` type alias to avoid this in the future 🤔

PR #306, opened 2022-04-27
Add jump-if-not-zero instruction.

### LINKED ISSUE #305: Add 18-bit conditional jump
Add a jump-if-not-zero-immediate instruction that implicitly compares against the `$zero` register, allowing for 18 bits of immediate jump address.

adlerjohn:
Fixes #305


mohammadfawaz:
Thanks! Looks good to me. I'll let @Voxelot make the final approval.

Voxelot (approved):

PR #314, opened 2022-05-12
Specify deposit IDs

pixelcircuits:
Updated deposit ID compute process

### DIFF
--- a/specs/protocol/identifiers.md
+++ b/specs/protocol/identifiers.md
@@ -4,8 +4,8 @@
 - [Contract ID](#contract-id)
 - [UTXO ID](#utxo-id)
   - [Coin ID](#coin-id)
-  - [Asset ID](#asset-id)
   - [Deposit ID](#deposit-id)
+    - [Deposit Asset ID](#deposit-asset-id)
   - [Fee ID](#fee-id)
 
 This document defines how to compute unique identifiers.
@@ -24,13 +24,14 @@ For a transaction of type `TransactionType.Create`, `tx`, the contract ID is `sh
 
 Is represented as an _outpoint_: a pair of [transaction ID](#transaction-id) as `byte[32]` and output index as a `uint8`.
 
-### Asset ID
-
-The ID of an asset is computed as the [hash](./cryptographic_primitives.md#hashing) of the token address `byte[32]` and token precision `uint8`: `hash(byte[32] ++ uint8)`. The address value is serialized as a byte array of length 32 left-padded with zeroes, and all other value types are serialized according to the standard [transaction serialization](./tx_format.md#transaction).
 
 ### Deposit ID
 
-The ID of a deposit is computed as the [hash](./cryptographic_primitives.md#hashing) of the [Asset ID](#asset-id) `byte[32]`, recipient address `byte[32]` and deposit nonce `uint64`: `hash(byte[32] ++ byte[32] ++ uint64)`. All value types are serialized according to the standard [transaction serialization](./tx_format.md#transaction).
+The ID of a deposit is computed as the [hash](./cryptographic_primitives.md#hashing) of the [Deposit Asset ID](#deposit-asset-id) `byte[32]`, recipient address `byte[32]`, amount `uint64` and deposit nonce `uint64`: `hash(byte[32] ++ byte[32] ++ uint64 ++ uint64)`. All value types are serialized according to the standard [transaction serialization](./tx_format.md#transaction).
+
+#### Deposit Asset ID
+
+The ID of a deposit asset is computed as the [hash](./cryptographic_primitives.md#hashing) of the token address `byte[32]` and token precision `uint8`: `hash(byte[32] ++ uint8)`. The address value is serialized as a byte array of length 32 left-padded with zeroes, and all other value types are serialized according to the standard [transaction serialization](./tx_format.md#transaction).
 
 ### Fee ID
 
--- a/specs/protocol/identifiers.md
+++ b/specs/protocol/identifiers.md
@@ -25,7 +25,7 @@ Is represented as an _outpoint_: a pair of [transaction ID](#transaction-id) as 
 
 ### Deposit ID
 
-The UTXO ID of a deposit is computed as the [hash](./cryptographic_primitives.md#hashing) of the depoitor `address`, token `address`, token precision `u8` and deposit nonce `u256`.
+The UTXO ID of a deposit is computed as the [hash](./cryptographic_primitives.md#hashing) of the depoitor address `byte[20]`, token address `byte[20]`, token precision `uint8` and deposit nonce `byte[32]`: `hash(byte[20] ++ byte[20] ++ uint8 ++ byte[32])`. Address values are serialized as a byte array of length 20 and all other value types are serialized according to the standard [transaction serialization](./tx_format.md#transaction).
 
 ### Fee ID
 
--- a/specs/protocol/identifiers.md
+++ b/specs/protocol/identifiers.md
@@ -25,7 +25,7 @@ Is represented as an _outpoint_: a pair of [transaction ID](#transaction-id) as 
 
 ### Deposit ID
 
-The UTXO ID of a deposit is computed as the [hash](./cryptographic_primitives.md#hashing) of TODO.
+The UTXO ID of a deposit is computed as the [hash](./cryptographic_primitives.md#hashing) of the depoitor `address`, token `address`, token precision `u8` and deposit nonce `u256`.
 
 ### Fee ID
 
--- a/specs/protocol/identifiers.md
+++ b/specs/protocol/identifiers.md
@@ -25,7 +25,7 @@ Is represented as an _outpoint_: a pair of [transaction ID](#transaction-id) as 
 
 ### Deposit ID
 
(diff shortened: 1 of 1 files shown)


pixelcircuits:
Not sure exactly what we want to make up the deposit ID. Technically the `depositNonce` would be enough, but I figured it might be useful to hash in other data bits to be able to more easily prove their existence if not immediately accessible in smart contracts / scripts.

Also, please advise where else this would need to be specified. I see #103 added to the tx_format part of the protocol to specify withdrawal outputs, but I'm not sure what the parallel would be for deposits.

@adlerjohn @simonr0204

specs/protocol/identifiers.md
```diff
@@ -25,7 +25,7 @@ Is represented as an _outpoint_: a pair of [transaction ID](#transaction-id) as
 
 ### Deposit ID
 
-The UTXO ID of a deposit is computed as the [hash](./cryptographic_primitives.md#hashing) of TODO.
+The UTXO ID of a deposit is computed as the [hash](./cryptographic_primitives.md#hashing) of the depoitor `address`, token `address`, token precision `u8` and deposit nonce `u256`.
```
adlerjohn, line 28: 1. There's no notion of Solidity types in the FuelVM. Specifically, the specs only use `uint8`, `uint16`, uint32`, uint64`, `byte[32]`, and `byte[]`.
2. You also have to specify how these are serialized (you can point to the canonical serialization format): https://github.com/FuelLabs/fuel-specs/blob/84cff9a590a28d7c5e6cb62d91b52e8e30c6ef67/specs/protocol/tx_format.md#transaction.

adlerjohn (commented):
You can leave specifying a deposit state element (similar to a UTXO) to a future PR, and only specify the actual identifier for a deposit in this PR.

specs/protocol/identifiers.md
pixelcircuits: I've updated with references to serialization processes. Does it make sense to hash all of these values though? Really only the deposit nonce is enough to make it unique, but we'd need to add some sort of extra salt to avoid collision of common hashes (hash(0x0001), hash(0x0002), etc.) so I can see it as being ok. It's also useful as a tool to prove the params of a deposit given it's ID

adlerjohn: It does make sense so that the deposit is self-verifying. Using nonces alone to represent anything other than ordering is actually pretty dangerous due to re-orgs and frontrunning: if someone can manipulate the order of txs to make them different than you expect, then they can convince you of something invalid that you have no obvious way of knowing is invalid.

adlerjohn: As for domain separation...there's an issue for that! #201. You could leave doing that to a follow-up PR.

specs/protocol/identifiers.md
```diff
@@ -25,7 +25,7 @@ Is represented as an _outpoint_: a pair of [transaction ID](#transaction-id) as
 
 ### Deposit ID
 
-The UTXO ID of a deposit is computed as the [hash](./cryptographic_primitives.md#hashing) of TODO.
+The UTXO ID of a deposit is computed as the [hash](./cryptographic_primitives.md#hashing) of the depoitor address `byte[20]`, token address `byte[20]`, token precision `uint8` and deposit nonce `byte[32]`: `hash(byte[20] ++ byte[20] ++ uint8 ++ byte[32])`. Address values are serialized as a byte array of length 20 and all other value types are serialized according to the standard [transaction serialization](./tx_format.md#transaction).
```
adlerjohn, line 28:
```suggestion
The UTXO ID of a deposit is computed as the [hash](./cryptographic_primitives.md#hashing) of the depositor address `byte[20]`, token address `byte[20]`, token precision `uint8` and deposit nonce `byte[32]`: `hash(byte[20] ++ byte[20] ++ uint8 ++ byte[32])`. Address values are serialized as a byte array of length 20 and all other value types are serialized according to the standard [transaction serialization](./tx_format.md#transaction).
```

adlerjohn (commented):
Hmm. Lots of stuff in the FuelVM works best at word alignment (i.e. 8 bytes). I would instead use 32 bytes for addresses, and left-pad with zeroes (which is what https://github.com/FuelLabs/sway/pull/1452) does. This would also be forwards-compatible with the proposed address extension: https://ethereum-magicians.org/t/thoughts-on-address-space-extension-ase/6779

Also, is there a need for 32 byte nonces? Would prefer 8 bytes (i.e. `uint64`) since it'll just be easier to deal with in the FuelVM.

pixelcircuits:
Changed address to be 32 bytes left-padded with zeroes and deposit nonce reduced to 64bits (still plenty)

adlerjohn (approved):

adlerjohn:
Looping in @simonr0204 and @rakita for reviews from the perspective of the contract and the relayer, respectively.

simonr0204:
How are we defining `precision` ? For the current ethereum bridges it's relative to 18 decimal places, since ether and many ERC20 tokens have 18 digits of precision. 

Do we want to be stuck with this definition for non-evm deployments ?

Also - is the address the _depositor_  (on L1) , or the Fuel address receiving the deposit?

adlerjohn:
Good questions. The precision here should be against units, e.g. wei for Ethereum. So a precision of `3` would mean 1,000 wei. Does that work?

As for address, it should be the receiving address on Fuel, which is 32 bytes so no padding necessary. The sending address is irrelevant for the Fuel chain.

simonr0204:
Makes sense. I'll update the contracts to reflect.

For the address, `depositor` may be a bit misleading then, as it sounds like it's referring to the L1 address initiating the deposit. Is there a better word? Receiver? Beneficiary?

pixelcircuits:
Maybe we refer to `depositor` as `recipient` instead?

adlerjohn:
Recipient ain't bad

adlerjohn (approved):

rakita:
> Looping in @simonr0204 and @rakita for reviews from the perspective of the contract and the relayer, respectively.

Relayer does not care what is inside of hash, we just need some kind of id, only having nonce would be sufficient enough for our side. 

Regarding deposits, in fuel ecosystem we don't have `u256`, we are using `u64` as an amount, so it is best to restrict max deposit amount to `u64::MAX`. 

I am not exactly getting how precision is going to work inside FuelVm, what is the idea behind it is this some data that is going to be used by VM or something like additional info that is tied to coin?

specs/protocol/identifiers.md
rakita: Deposits will not be able to be used until they are finalized, I wouldn't recommend SDK to include it inside the next tx until we are sure that finalization is done. 
We can make that contract sends us nonce as id, while we mutate this into the id (hash of all data) that we want to be presented to users/sdk?

adlerjohn:
> I am not exactly getting how precision is going to work inside FuelVm, what is the idea behind it is this some data that is going to be used by VM or something like additional info that is tied to coin?

The FuelVM won't actually use the precision. Token contracts can be provided the preimage to the ID to extract the precision, then use that to mint an appropriate number of Fuel-tokens.

specs/protocol/identifiers.md
```diff
@@ -25,7 +25,7 @@ Is represented as an _outpoint_: a pair of [transaction ID](#transaction-id) as
 
 ### Deposit ID
 
-The UTXO ID of a deposit is computed as the [hash](./cryptographic_primitives.md#hashing) of TODO.
+The UTXO ID of a deposit is computed as the [hash](./cryptographic_primitives.md#hashing) of the recipient address `byte[32]`, token address `byte[32]`, token precision `uint8` and deposit nonce `uint64`: `hash(byte[32] ++ byte[32] ++ uint8 ++ uint64)`. Address values are serialized as a byte array of length 32 left-padded with zeroes, and all other value types are serialized according to the standard [transaction serialization](./tx_format.md#transaction).
```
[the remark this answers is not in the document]
rakita: Just reposting this here so that it is not lost in the comments: Deposits will not be able to be used until they are finalized, I wouldn't recommend SDK to include it inside the next tx until we are sure that finalization is done.
We can make that contract sends us nonce as id, while we mutate this into the id (hash of all data) that we want to be presented to users/sdk?

specs/protocol/identifiers.md
```diff
@@ -23,9 +24,13 @@ For a transaction of type `TransactionType.Create`, `tx`, the contract ID is `sh
 
 Is represented as an _outpoint_: a pair of [transaction ID](#transaction-id) as `byte[32]` and output index as a `uint8`.
 
+### Asset ID
```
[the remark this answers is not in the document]
pixelcircuits: sure!

specs/protocol/identifiers.md
```diff
@@ -23,9 +24,13 @@ For a transaction of type `TransactionType.Create`, `tx`, the contract ID is `sh
...
+
+The ID of an asset is computed as the [hash](./cryptographic_primitives.md#hashing) of the token address `byte[32]` and token precision `uint8`: `hash(byte[32] ++ uint8)`. The address value is serialized as a byte array of length 32 left-padded with zeroes, and all other value types are serialized according to the standard [transaction serialization](./tx_format.md#transaction).
+
 ### Deposit ID
 
-The UTXO ID of a deposit is computed as the [hash](./cryptographic_primitives.md#hashing) of TODO.
+The ID of a deposit is computed as the [hash](./cryptographic_primitives.md#hashing) of the [Asset ID](#asset-id) `byte[32]`, recipient address `byte[32]` and deposit nonce `uint64`: `hash(byte[32] ++ byte[32] ++ uint64)`. All value types are serialized according to the standard [transaction serialization](./tx_format.md#transaction).
```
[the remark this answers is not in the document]
pixelcircuits: I don't see any harm in being complete and adding it in case it becomes useful later

Issue #298, opened 2022-03-27
Widen tx field counts to 64 bits


adlerjohn:
8 and 16 bits are used throughout, but are too small.

AlicanC:
One issue with u64 for lengths & counts is that we can't represent these easily in JS. A recent update I've made got rid of a lot of ugliness because we were finally using `number`s for sizes <64:

```ts
const myArray = [1, 2, 3];

// Before with u16 (or would be the same with u32)
new ArrayEncoder(new NumberCoder('u64'), BigInt(myArray.length)).encode(myArray);

// After with u64
new ArrayEncoder(new NumberCoder('u64'), myArray.length).encode(myArray);
```

This isn't an issue for non-count non-size stuff since they can exist as they are, but for counts/sizes we need to go from number to bigint constantly because JS always uses number for lengths.

Can we keep these as u32s?

adlerjohn:
In practice, most of these fields won't exceed the limit of a `u32` even if their _type_ is `u64`. There will be consensus rules placing an upper bound on each of these fields that will be well below the maximum `u64` (or even the `maximum `u32`) value. Is that sufficient @AlicanC, or does the _type_ need to be `u32`?

AlicanC:
If the type is u32 the coding infra will just handle it naturally. Otherwise we will have to introduce a concept of `things that are defined as u64 but should be treated as u32 for convenience` and have code to handle that.

MitchTurner:
I think `u16` should be enough. Can we close this @xgreenx ?

xgreenx:
Yeah, `u16` sound good for 1 second blocks for now

PR #318, opened 2022-05-20
New messaging specification (replacing old deposit/withdrawal spec)

pixelcircuits:
Changes to keep track of messages going in and out

Implementation:

- [ ] fuel-tx: https://github.com/FuelLabs/fuel-tx/issues/145


rakita:
Just a comment for clarity, in my head, the message could contain:
* `GatewayIdFrom` -> Registered gateway from this side of the bridge.
* `GatewayIdTo` -> Registered gateway from the other side of the bridge.
* `Data` -> whatever it is. It is gateway specific
* `Nonce` -> so that we don't have a replay attack. 

The Data field contains all needed information as it depends on what we want for the gateway to do, for erc20 example, we will need info on who is the new owner of the token, token id, and amount.

MessageId can be hash over message fields.

It feels like we could remove `GatewayFrom` and use just `GatewayIdTo` but we need to think about how to do the registration procedure and unique mapping in the router.

pixelcircuits:
In the same style as Arbitrum/Optimism, I tried to keep the core message object as generic as possible to facilitate more than gateways. For example, a gateway deposit from Fuel to Ethereum would work like this:

- `EVM address`: Ethereum side gateway address (to)
- `FuelVM address`: Fuel side gateway address (from)
- `Message ABI`: The call to execute on the "to" contract address which in this example is the Ethereum side gateway `abi.encodeWithSignature("finalizeDeposit(uint256,uint256,uint256,uint256,uint256,bytes)", l1Token, l2Token, from, to, amount, data)`
- `Nonce`: Enforces ID uniqueness

I also imagined that there wouldn't be any onchain router and it would be up to the front end UI to help a user pick a gateway contract for a particular asset. Similar to how uniswap uses their off chain token lists to help a user swap with the correct token address they're looking for.

I'm currently in the process of diagraming everything up to be scrutinized before we start writing contracts.

pixelcircuits:
Who else should be asked to review these messaging spec updates?

specs/protocol/tx_validity.md
```diff
@@ -84,14 +84,14 @@ For each asset ID `asset_id` in the input and output set:
...
-        if input.type == InputType.Coin and input.asset_id == col:
+        if (input.type == InputType.Coin and input.asset_id == col) or (input.type == InputType.Message and col == 0):
             total += input.amount
     return total
 
 def sum_outputs(tx, col) -> int:
     total: int = 0
     for output in tx.outputs:
-        if (output.type == OutputType.Coin or output.type == OutputType.Withdrawal) and output.asset_id == col:
+        if (output.type == OutputType.Coin and output.asset_id == col) or (output.type == OutputType.Message and col == 0):
```
Voxelot, line 94: wait, why would we include the amount of message outputs for `col == 0` if this method is calculating the total for `col == x`?

Voxelot: This means if I sent some USDC in the same tx I withdraw, it would add the eth/fuel amount of the withdraw message to the USDC amount.

pixelcircuits: Color 0 is the base asset right? An OutputMessage with an amount value is effectively a withdrawal/burn on the base asset and must be accounted for.

pixelcircuits: The amount field is only used to indicate the base asset (ETH) being sent along with the message to the recipient. Like how you can send a value as part of an EVM call

Voxelot: Ah ok, I think I misread as `output.asset_id == 0`. So this is just special casing output messages since they don't have variable asset_id's?

pixelcircuits: Correct. The amount value is always for asset 0

Voxelot: cool, disregard my comment then

pixelcircuits:
Should some sort of receipt exist for every time the base asset is minted and/or burned from messages going in and out?

specs/protocol/identifiers.md
```diff
@@ -24,14 +23,9 @@ For a transaction of type `TransactionType.Create`, `tx`, the contract ID is `sh
...
-
-The ID of a deposit asset is computed as the [hash](./cryptographic_primitives.md#hashing) of the token address `byte[32]` and token precision `uint8`: `hash(byte[32] ++ uint8)`. The address value is serialized as a byte array of length 32 left-padded with zeroes, and all other value types are serialized according to the standard [transaction serialization](./tx_format.md#transaction).
+The ID of a message is computed as the [hash](./cryptographic_primitives.md#hashing) of the sender address `byte[32]`, recipient address `byte[32]`, the message data (either [EVM abi](https://docs.soliditylang.org/en/v0.8.13/abi-spec.html) or [FuelVM abi](./abi.md)/data) `byte[]`, the amount being sent with the message `uint64`, the message nonce `uint64` and the message owner (either address or predicate hash) `byte[32]`: `hash(byte[32] ++ byte[32] ++ byte[] ++ uint64 ++ uint64 ++ byte[32])`. The address values are serialized as a byte array of length 32 left-padded with zeroes, and all other value types are serialized according to the standard [transaction serialization](./tx_format.md#transaction). If the owner component is not applicable then a 32 byte long array of zeros should be used.
```
adlerjohn, line 28: There should be a data length too, no?
```suggestion
The ID of a message is computed as the [hash](./cryptographic_primitives.md#hashing) of:

1. the sender address `byte[32]`,
1. recipient address `byte[32]`,
1. the message data `byte[]`,
1. the amount being sent with the message `uint64`,
1. the message nonce `uint64`,
1. the message owner (either address or predicate hash)

`byte[32]`: `hash(byte[32] ++ byte[32] ++ byte[] ++ uint64 ++ uint64 ++ byte[32])`. The address values are serialized as a byte array of length 32 left-padded with zeroes, and all other value types are serialized according to the standard [transaction serialization](./tx_format.md#transaction). If the owner component is not applicable then a 32 byte long array of zeros should be used.
```

specs/protocol/tx_format.md
```diff
@@ -178,7 +180,7 @@ enum  InputType : uint8 {
 | name   | type                                                              | description    |
 |--------|-------------------------------------------------------------------|----------------|
 | `type` | [InputType](#inputtype)                                           | Type of input. |
-| `data` | One of [InputCoin](#inputcoin) or [InputContract](#inputcontract) | Input data.    |
+| `data` | One of [InputCoin](#inputcoin), [InputContract](#inputcontract) or [InputMessage](#inputmessage) | Input data.    |
```
adlerjohn, line 183:
```suggestion
| `data` | One of [InputCoin](#inputcoin), [InputContract](#inputcontract), or [InputMessage](#inputmessage) | Input data.    |
```

specs/protocol/tx_format.md
```diff
@@ -9,11 +9,12 @@
 - [Input](#input)
   - [InputCoin](#inputcoin)
   - [InputContract](#inputcontract)
+  - [InputMessage](#inputmessage)
 - [OutputType](#outputtype)
 - [Output](#output)
   - [OutputCoin](#outputcoin)
   - [OutputContract](#outputcontract)
-  - [OutputWithdrawal](#outputwithdrawal)
+  - [OutputMessage](#outputmessage)
```
adlerjohn, line 17: Hmmm. Should we have an output message and an output message variable? Would be nice for users to potentially add messages to their tx without a script. Not strictly necessary.

specs/protocol/tx_format.md
```diff
@@ -239,13 +241,40 @@ Note: when verifying a predicate, `txID`, `outputIndex`, `balanceRoot`, `stateRo
...
 
+### InputMessage
+
+| name                  | type        | description                                                            |
+|-----------------------|-------------|------------------------------------------------------------------------|
+| `messageID`           | `byte[32]`  | The messageID as described [here](./identifiers.md#message-id).        |
+| `sender`              | `byte[32]`  | The address of the message sender.                                     |
+| `recipient`           | `byte[32]`  | The address of the message recipient.                                  |
+| `dataLength`          | `uint16`    | Length of message data, in bytes.                                      |
+| `data`                | `byte[]`    | The message data or [abi encoded](./abi.md) call to execute.           |
```
adlerjohn, line 252: We don't really need Solidity ABI encoding here I think? It would be Fuel encoded data, but that's application logic. The protocol rules don't care about what the data is.
```suggestion
| `data`                | `byte[]`    | The message data .           |
```

adlerjohn, line 252: Move this to the back. Note that fixed-length field are first, then all variable-length fields. This is required to be able to read fixed-length fields in constant time.

specs/protocol/tx_format.md
```diff
@@ -239,13 +241,40 @@ Note: when verifying a predicate, `txID`, `outputIndex`, `balanceRoot`, `stateRo
...
+| `predicateData`       | `byte[]`    | Predicate input data (parameters).                                     |
+
+Transaction is invalid if:
+
+- `witnessIndex >= tx.witnessesCount`
+- `predicateLength > MAX_PREDICATE_LENGTH`
+- `predicateDataLength > MAX_PREDICATE_DATA_LENGTH`
+- If `predicateLength > 0`; the computed predicate root (see below) is not equal `owner`
+
+Note: the predicate root is computed identically to the contract ID, [here](./identifiers.md#contract-id).
```
adlerjohn, line 269: #321
```suggestion
The predicate root is computed identically to the contract root, used to compute the contract ID, [here](./identifiers.md#contract-id).
```

specs/protocol/tx_format.md
```diff
@@ -294,15 +323,23 @@ The balance root `balanceRoot` is the root of the [SMT](./cryptographic_primitiv
 
 The state root `stateRoot` is the root of the [SMT](./cryptographic_primitives.md#sparse-merkle-tree) of storage slots. Each storage slot is a `byte[32]`, keyed by a `byte[32]`.
 
-### OutputWithdrawal
+### OutputMessage
+
+| name                  | type        | description                                                             |
+|-----------------------|-------------|-------------------------------------------------------------------------|
+| `messageID`           | `byte[32]`  | The messageID as described [here](./identifiers.md#message-id).         |
```
adlerjohn, line 330: Wait, why does this have a message ID? Isn't that only for inputs?

specs/vm/opcodes.md
```diff
@@ -76,6 +76,7 @@
   - [SWWQ: State write 32 bytes](#swwq-state-write-32-bytes)
   - [TR: Transfer coins to contract](#tr-transfer-coins-to-contract)
   - [TRO: Transfer coins to output](#tro-transfer-coins-to-output)
+  - [SMO: Send Message to output](#smo-send-message-to-output)
```
adlerjohn, line 79: Alphabetize (will require moving the actual instruction definition).

specs/protocol/identifiers.md
pixelcircuits: Not necessarily. In the solidity contracts I haven't been using an explicitly defined length. It technically reduces gas cost for hashing, but not by a lot if we find it extra useful somewhere else

specs/protocol/tx_format.md
pixelcircuits: We could. At the very least maybe we make the `SMO` opcode available to transaction scripts. I also feel like messages output from smart contracts should be the default so we should just call "OutputMessageVariable" `OutputMessage` and if we make an option for user to sign messages without a script we call it "OutputMessageDirect" or "OutputMessageImmediate" or something

specs/protocol/tx_format.md
pixelcircuits: It's referring to Fuels abi with the thought being that most messages will be going through a standard relayer contract where the message data will be the abi for calling a Fuel contract function. But I agree that this is application specific and shouldn't be in the spec.

specs/protocol/tx_format.md
pixelcircuits: Right now, the message ID for outputs is what's placed in the binary merkle tree that gets committed to Ethereum. It's these message IDs that the Inbox uses to ensure a message is never relayed more than once

specs/protocol/tx_validity.md
```diff
@@ -84,14 +84,14 @@ For each asset ID `asset_id` in the input and output set:
 def sum_inputs(tx, col) -> int:
     total: int = 0
     for input in tx.inputs:
-        if input.type == InputType.Coin and input.asset_id == col:
+        if (input.type == InputType.Coin and input.asset_id == col) or (input.type == InputType.Message and col == 0):
```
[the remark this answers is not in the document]
pixelcircuits: because the `input.amount` field on an `InputMessage` only counts towards the base asset (color 0)

specs/protocol/tx_format.md
adlerjohn: We can always add the direct message in a future PR if it's needed, since it's conceptually a simple change. Let's keep this one simple.

specs/protocol/tx_format.md
adlerjohn: Ah I see. In that case, is there really a need for the rest of the fields? Those can be placed in the receipt, rather than in the output.

specs/protocol/tx_validity.md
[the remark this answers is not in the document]
adlerjohn: Ahhhh I see now, I misread.

specs/protocol/identifiers.md
adlerjohn: We probably want the data length otherwise there's no way to know how long it is. I guess there's only one variable-length field, so you could infer it solely by looking at the total date length, but that seems kind of brittle.

specs/protocol/tx_format.md
pixelcircuits: I thought we weren't doing receipts for messages because all the data is already in the output? The fields needs to be somewhere that's easy for a user to pull from a client so anyone can then relay the messages on L1

specs/vm/instruction_set.md
```diff
@@ -1386,6 +1387,42 @@ Cease VM execution and revert script effects. After a revert:
...
+- `tx.outputs[$rC].type != OutputType.Message`
+- `tx.outputs[$rC].recipient != 0 || tx.outputs[$rC].sender != 0`
+
+In an external context, decrease `MEM[balanceOfStart(0), 8]` by `$rD`. In an internal context, decrease asset ID 0 balance of output with contract ID `MEM[$fp, 32]` by `$rD`. Then set:
+
+- `tx.outputs[$rC].recipient = MEM[$rA, 32]`
+- `tx.outputs[$rC].sender = MEM[$fp, 32]`
+- `tx.outputs[$rC].callABI = MEM[$rA + 32, $rB]`
+- `tx.outputs[$rC].amount = $rD`
+- `tx.outputs[$rC].nonce = ++messageOutputNonce`
```
rakita, line 1420: Is this a global `messageOutputNonce` variable? as transactions can be executed in parallel this can become a synchronization problem. I propose that nonce be left zero in instruction and in the loop, after block is executed, we transverse all transactions and set the nonce, we already need to do that loop to create merkle tree for all MessageOutputs. And globals are not something desirable, maybe we can somehow mix fuel block height and local block Nonce?

pixelcircuits: yeah, it was imagined as a global variable. Setting the nonce later is ok, but that also means we can't calculate the messageID until this is filled in. It can be whatever as long as it's available when an app or user looks to a processed transaction to find the outputs (to get the messageID and lookup the merkle proof for eventual relaying). What exactly do you think it should be?

rakita: Introducing two variables, `block_height` u32 and `nonce` (that is local to the block) u32. It will break the symmetry that we have but will remove the need for global vars. MessageId needs to be unique and height/local_nonce do that. wdyt?

pixelcircuits: So this would help remove the need for a global variable, but does it solve the problem of running txs in parallel? What if we treat the nonce like `hash(tx ++ OutputMessageIndex)` where `tx` is some sort of unique tx identifier and `OutputMessageIndex` is the index of the output message in the list of the transactions output?

rakita: a solution that would allow running txs in parallel is to not set nonce in runtime in `SMO` but in post-processing when block is already created.
`hash(tx++OutputMessageIndex)` sound nice, it is unique and similar to [UtxoId](https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/identifiers.md#utxo-id) but this will increase our nonce size to 256bits if it is keccake hash.

adlerjohn: Do we actually need the nonce to be unique? Why is `txId ++ OutputMessageIndex` (no hashing!) not a sufficiently unique identifier?

rakita: It is sufficient, it is just a matter of choosing.

pixelcircuits: So with the nonce being expanded to 256bits now, would `txId + OutputMessageIndex` be something that can be computed at the time of execution (no need for a second pass after building a block)

specs/protocol/tx_format.md
pixelcircuits: I've now moved most of the OutputMessage data into a receipt. Data length is capped at 2**16 or 64kB

specs/vm/instruction_set.md
rakita: TxId+OutputIndex is good and i assume that they can be set and checked at runtime, let us go with it. 256bit for TxId and u8 for the index as MAX_OUTPUTS is 8: https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/tx_format.md#constants
A second pass is needed for merkalization of message ids. I will resolve this comments if it is okay, as consern with global nonce was mitigated.

adlerjohn: > would txId + OutputMessageIndex be something that can be computed at the time of execution

Actually yes! The tx ID is computed off of the non-malleable parts only (technically, the malleable parts are set to zero, but whatever). And it's actually available during VM execution at memory address zero.

adlerjohn: Ah so the max inputs/outputs were changed to 255 in the implementation, but never updated in the specs. And there was some discussion to change the type to `u64`, lost to the void of time.

specs/vm/instruction_set.md
```diff
@@ -1386,6 +1387,53 @@ Cease VM execution and revert script effects. After a revert:
...
+- `$rA + $rB + 32` overflows
+- `$rA + 32 > VM_MAX_RAM`
+- `$rA + $rB + 32 > VM_MAX_RAM`
+- `$rC > tx.outputsCount`
+- In an external context, if `$rD > MEM[balanceOfStart(0), 8]`
+- In an internal context, if `$rD` is greater than the balance of asset ID 0 of output with contract ID `MEM[$fp, 32]`
+- `tx.outputs[$rC].type != OutputType.Message`
+- `tx.outputs[$rC].recipient != 0 || tx.outputs[$rC].sender != 0`
+
+Append a receipt to the list of receipts, modifying `tx.receiptsRoot`:
```
adlerjohn, line 1414: I think this receipt needs to be in the JSON ABI spec as well: https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md#receipt

specs/vm/instruction_set.md
```diff
@@ -1386,6 +1387,53 @@ Cease VM execution and revert script effects. After a revert:
...
+| name         | type          | description                                                               |
+|--------------|---------------|---------------------------------------------------------------------------|
+| `type`       | `ReceiptType` | `ReceiptType.MessageOut`                                                  |
+| `messageID`  | `byte[32]`    | The messageID as described [here](../protocol/identifiers.md#message-id). |
+| `sender`     | `byte[32]`    | The address of the message sender: `MEM[$fp, 32]`.                        |
+| `recipient`  | `byte[32]`    | The address of the message recipient: `MEM[$rA, 32]`.                     |
+| `dataLength` | `uint16`      | Length of message data, in bytes: `$rB - ($rA + 32)`.                     |
+| `amount`     | `uint64`      | Amount of base asset coins sent with message: `$rD`.                      |
+| `nonce`      | `byte[32]`    | The message nonce.                                                        |
+| `data`       | `byte[]`      | The message data or [abi encoded](https://docs.soliditylang.org/en/v0.8.13/abi-spec.html) call to execute: `MEM[$rA + 32, $rB]`.        |
```
adlerjohn, line 1425: I would remove the reference to Solidity here, since the Fuel protocol doesn't care what this data is.

specs/vm/instruction_set.md
```diff
@@ -1386,6 +1387,53 @@ Cease VM execution and revert script effects. After a revert:
...
+| Notes       |                                                                                     |
+
+Given helper `balanceOfStart(asset_id: byte[32]) -> uint32` which returns the memory address of the remaining free balance of `asset_id`, or panics if `asset_id` has no free balance remaining.
+
+Panic if:
+
+- `$rA + 32` overflows
+- `$rA + $rB + 32` overflows
+- `$rA + 32 > VM_MAX_RAM`
+- `$rA + $rB + 32 > VM_MAX_RAM`
```
adlerjohn, line 1407: There should also be a check that `$rB` does not exceed `MEM_MAX_ACCESS_SIZE`, the maximum atomic memory access size.

specs/vm/instruction_set.md
adlerjohn, line 1425: Rather than `data`, this should be `digest`, similar to the `LOGD` and `RETD` receipts. Essentially, this guarantees that computing the receipt hash is constant time. The actual data is collected by the implementation and available in the returned receipt implementations-side. For example: https://github.com/FuelLabs/fuel-specs/blob/fcfe18a585a465c750fd939b0fd85a30255d6c00/specs/protocol/abi.md#logdata-receipt

specs/vm/instruction_set.md
pixelcircuits: Isn't this implied? I don't see any reference to the abi for any other opcode that creates receipts. Should we create a separate issue to add reference to the abi for all opcodes that mention receipts?

adlerjohn: If you look at the above link, you'll see each of the receipt types have a JSON definition. This is unrelated to the instruction set, hence why other instructions don't talk about what the JSON receipts look like.

adlerjohn: As for whether instructions that emit receipts should also link to the relevant sections in the JSON ABI spec...probably yes!

specs/vm/instruction_set.md
pixelcircuits: The `LOGD` and `RETD` include both a `digest` and the full `data`. I'll be sure to add a `digest` to the receipt, but I'm confused as to what you mean by "receipt hash"

adlerjohn: To clarify: the _canonical_ `LOGD` receipt (the thing that's in the instruction set file, here) only includes a digest (similar for `RETD`): https://github.com/FuelLabs/fuel-specs/blob/fcfe18a585a465c750fd939b0fd85a30255d6c00/specs/vm/instruction_set.md#logd-log-data-event. The _JSON_ `LOGD` receipt includes both the digest and data (i.e. preimage to the digest): https://github.com/FuelLabs/fuel-specs/blob/fcfe18a585a465c750fd939b0fd85a30255d6c00/specs/protocol/abi.md#logdata-receipt

By "receipt hash" I actually meant "hashing a leaf in the receipts tree." If leaves only contain digests (they do), then they're constant-sized and that makes it easier to do some fraud proof cases and also serve light client proofs. Sometimes you don't care about the data but only that a receipt happened. If the data isn't part of the canonical receipt then you can still get a proof to the canonical receipt without having to download all the data.

specs/protocol/tx_format.md
```diff
@@ -230,13 +233,40 @@ Note: when verifying a predicate, `txID`, `outputIndex`, `balanceRoot`, `stateRo
...
 Note: when executing a script, `txID`, `outputIndex`, `balanceRoot`, and `stateRoot` are initialized to the transaction ID, output index, amount, and state root of the contract with ID `contractID`, and `txPointer` is initialized to zero.
 
+### InputMessage
+
+| name                  | type        | description                                                            |
+|-----------------------|-------------|------------------------------------------------------------------------|
+| `messageID`           | `byte[32]`  | The messageID as described [here](./identifiers.md#message-id).        |
+| `sender`              | `byte[32]`  | The address of the message sender.                                     |
+| `recipient`           | `byte[32]`  | The address of the message recipient.                                  |
+| `dataLength`          | `uint16`    | Length of message data, in bytes.                                      |
```
adlerjohn, line 243: I would put this right before `predicateLength`.

specs/protocol/tx_format.md
```diff
@@ -230,13 +233,40 @@ Note: when verifying a predicate, `txID`, `outputIndex`, `balanceRoot`, `stateRo
...
+| `nonce`               | `byte[32]`  | The message nonce.                                                     |
+| `owner`               | `byte[32]`  | Owning address or predicate hash.                                      |
+| `witnessIndex`        | `uint8`     | Index of witness that authorizes spending the coin.                    |
+| `predicateLength`     | `uint16`    | Length of predicate, in instructions.                                  |
+| `predicateDataLength` | `uint16`    | Length of predicate input data, in bytes.                              |
+| `data`                | `byte[]`    | The message data.                                                      |
+| `predicate`           | `byte[]`    | Predicate bytecode.                                                    |
+| `predicateData`       | `byte[]`    | Predicate input data (parameters).                                     |
+
+Transaction is invalid if:
```
adlerjohn, line 254: Hmmm. There should be a bound here on the data length. Maybe make a new consensus parameter, `MAX_MESSAGE_DATA_LENGTH`.

specs/vm/instruction_set.md
```diff
@@ -1386,6 +1387,55 @@ Cease VM execution and revert script effects. After a revert:
...
+| Operation   | ```outputmessage(MEM[$fp, 32], MEM[$rA, 32], MEM[$rA + 32, $rB], $rD, $rC);```      |
+| Syntax      | `smo $rA, $rB, $rC, $rD`                                                            |
+| Encoding    | `0x00 rA rB rC rD`                                                                  |
+| Notes       |                                                                                     |
+
+Given helper `balanceOfStart(asset_id: byte[32]) -> uint32` which returns the memory address of the remaining free balance of `asset_id`, or panics if `asset_id` has no free balance remaining.
+
+Panic if:
+
+- `$rA + 32` overflows
```
adlerjohn, line 1404: This is redundant with the below condition, no?

specs/vm/instruction_set.md
```diff
@@ -1386,6 +1387,55 @@ Cease VM execution and revert script effects. After a revert:
...
+- `$rA + $rB + 32` overflows
+- `$rA + 32 > VM_MAX_RAM`
```
adlerjohn, line 1406: Ditto, redundant

specs/vm/instruction_set.md
```diff
@@ -1386,6 +1387,55 @@ Cease VM execution and revert script effects. After a revert:
...
+- `$rA + $rB + 32 > VM_MAX_RAM`
+- `$rB > MEM_MAX_ACCESS_SIZE`
+- `$rB > 65535`
```
adlerjohn, line 1409: Where does `65535` come from? Is it to fit into a `u16`? If so, then that should be captured by `MEM_MAX_ACCESS_SIZE`.

specs/vm/instruction_set.md
```diff
@@ -1386,6 +1387,55 @@ Cease VM execution and revert script effects. After a revert:
...
+- `$rC > tx.outputsCount`
+- In an external context, if `$rD > MEM[balanceOfStart(0), 8]`
+- In an internal context, if `$rD` is greater than the balance of asset ID 0 of output with contract ID `MEM[$fp, 32]`
+- `tx.outputs[$rC].type != OutputType.Message`
+- `tx.outputs[$rC].recipient != 0 || tx.outputs[$rC].sender != 0`
```
adlerjohn, line 1414: This looks old. It should be sufficient to check that amount is not zero. You should also add a check that `$rD` is not zero.

adlerjohn (commented):
Some minor things on checks, but other than that it's looking good.

specs/vm/instruction_set.md
pixelcircuits: MEM_MAX_ACCESS_SIZE is a u64. Is this incorrect in the spec?

specs/vm/instruction_set.md
pixelcircuits: This is to prevent overwriting an already set message output (Similar to TROs `tx.outputs[$rB].amount != 0`). `$rD` is the amount being sent, which can be zero (it's zero for the ERC20 gateway)

adlerjohn: Ahh you're correct, since you can send a non-zero amount with the message. Regardless, the fields here look like they're from the old version of message output.

specs/vm/instruction_set.md
adlerjohn: Yes, first because that's just the type of the value, second because of https://github.com/FuelLabs/fuel-specs/issues/298

specs/vm/instruction_set.md
pixelcircuits: oh yeah, only `recipient` and `amount` are on an output message

specs/vm/instruction_set.md
pixelcircuits: Ok, well maybe we should keep them different in case one changes but not the other. I'm talking about defining a `MESSAGE_MAX_DATA_SIZE` similar to `CONTRACT_MAX_SIZE`

adlerjohn: Yes, it could be `MESSAGE_MAX_DATA_LENGTH` described here https://github.com/FuelLabs/fuel-specs/pull/318#discussion_r895074789

adlerjohn: Ah nvm, got it, one's for input and the other's for the receipt.

specs/vm/instruction_set.md
```diff
@@ -1416,6 +1417,53 @@ Cease VM execution and revert script effects. After a revert:
...
+
+- `$rA + $rB + 32` overflows
+- `$rA + $rB + 32 > VM_MAX_RAM`
+- `$rB > MEM_MAX_ACCESS_SIZE`
+- `$rB > MESSAGE_MAX_DATA_SIZE`
+- `$rC > tx.outputsCount`
+- In an external context, if `$rD > MEM[balanceOfStart(0), 8]`
+- In an internal context, if `$rD` is greater than the balance of asset ID 0 of output with contract ID `MEM[$fp, 32]`
+- `tx.outputs[$rC].type != OutputType.Message`
+- `tx.outputs[$rC].recipient != 0`
```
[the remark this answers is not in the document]
adlerjohn: +1 to above. If the invariant is that recipient being zero means the output it unused, then it must be non-zero if the output is used. It would be rather ugly to enforce though since maybe some chains have something at address zero.

An alternative would be to include a `used` Boolean field, or something to that effect, that is set when the output it used..

specs/vm/instruction_set.md
```diff
@@ -1416,6 +1417,53 @@ Cease VM execution and revert script effects. After a revert:
...
+Append a receipt to the list of receipts, modifying `tx.receiptsRoot`:
+
+| name         | type          | description                                                               |
+|--------------|---------------|---------------------------------------------------------------------------|
+| `type`       | `ReceiptType` | `ReceiptType.MessageOut`                                                  |
+| `messageID`  | `byte[32]`    | The messageID as described [here](../protocol/identifiers.md#message-id). |
+| `sender`     | `byte[32]`    | The address of the message sender: `MEM[$fp, 32]`.                        |
+| `recipient`  | `byte[32]`    | The address of the message recipient: `MEM[$rA, 32]`.                     |
+| `amount`     | `uint64`      | Amount of base asset coins sent with message: `$rD`.                      |
+| `nonce`      | `byte[32]`    | The message nonce.                                                        |
```
adlerjohn, line 1453: Rather than nonce, it could be simpler to just use the output index, i.e. `$rC`.

specs/vm/instruction_set.md
```diff
@@ -1416,6 +1417,53 @@ Cease VM execution and revert script effects. After a revert:
...
+| `len`        | `uint16`      | Length of message data, in bytes: `$rB`.                                  |
+| `digest`     | `byte[32]`    | [Hash](#s256-sha-2-256) of `MEM[$rA + 32, $rB]`.                          |
+
+In an external context, decrease `MEM[balanceOfStart(0), 8]` by `$rD`. In an internal context, decrease asset ID 0 balance of output with contract ID `MEM[$fp, 32]` by `$rD`. Then set:
+
+- `tx.outputs[$rC].recipient = MEM[$rA, 32]`
+- `tx.outputs[$rC].amount = $rD`
+
+This modifies the `balanceRoot` field of the appropriate output(s).
```
adlerjohn, line 1462: Only a single contract output can be modified by the operation, so we can explicitly remove the plural.

```suggestion
This modifies the `balanceRoot` field of the appropriate output.
```

specs/vm/instruction_set.md
```diff
@@ -1416,6 +1417,53 @@ Cease VM execution and revert script effects. After a revert:
...
+MessageID is added to the `OutputMessage` merkle tree as part of block header.
+
+Append messageID to the global list of `OutputMessage`, modifying `OutputMessagesRoot`:
```
adlerjohn, line 1465: Maybe add a TODO here to further specify once the block header spec has progressed.
```suggestion
`messageID` is added to the `OutputMessage` Merkle tree as part of block header.

Append `messageID` to the global list of `OutputMessage`, modifying `OutputMessagesRoot`.
```

specs/vm/instruction_set.md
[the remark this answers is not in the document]
pixelcircuits: If a chain has something meaningful at the zero address, then the inbox contract on L1 could just interpret some designated address as really the zero address. Or there could be some forwarding contract on L1 that passes messages along to the zero address

specs/vm/instruction_set.md
pixelcircuits: It needs to be something unique to each message, or else someone malicious could force two messages to have the same `messageID`. I thought we were decided on `hash(txID ++ outputIndex)`. I'll add the concept of `OutputMessageNonce` to the identifiers section and link here

adlerjohn: Transaction IDs are unique across the entire history of the chain, so by construction `txID ++ outputIndex` is collision-free.

adlerjohn (commented):
I think the last thing that was discussed in the call was separating input and output messages, which should be a minor change. Other than that it's looking solid.

specs/vm/instruction_set.md
pixelcircuits: The nonce concept is still used when verifying merkle proofs as a way to reduce calldata (provide just a hash vs a byte32 + uint8). I'm leaning towards logging the full nonce rather than the bits to compute the nonce simply out of convenience when fetching the data required to prove a message on L1. Similar to how we log the full messageID even though it could just be computed from everything else.

adlerjohn: Hmm but the receipt and what's in the Merkle root of outbound messages to the EVM chain are actually independent. They can contain different values. Can you also clarify what specifically this nonce is? Is it just some unique identifier for the message? If so, how does it differ from message ID.

pixelcircuits: The nonce is here to ensure that each messageID is unique. The relay function on the L1 inbox looks like this: `relayMessage(bytes32 sender, bytes32 recipient, bytes32 nonce, uint64 amount, bytes data, BinaryMerkleProof merkleProof)`. The function hashes all values to figure out the `messageID` and then uses the provided merkle proof to verify the `messageID` is valid. If we didn't have the nonce, we'd have to pass in both `byte32 txID` and `uint8 outputIndex` requiring more calldata. The nonce is defined [Here](https://github.com/FuelLabs/fuel-specs/blob/pixelcircuits-refactor-depositwithdraw-to-flat-messaging/specs/protocol/identifiers.md#output-message-nonce)

adlerjohn: Ah got it. Can you link to that in here? Something like

```
The message [nonce](../protocol/identifiers.md#output-message-nonce).
```

specs/protocol/identifiers.md
```diff
@@ -24,14 +25,34 @@ For a transaction of type `TransactionType.Create`, `tx`, the contract ID is `sh
...
+1. recipient address as `byte[32]`,
+1. the [OutputMessage nonce](#output-message-nonce) as `byte[32]`,
+1. the amount being sent with the message as `uint64`,
+1. the message data as `byte[]`
+
+`hash(byte[32] ++ byte[32] ++ byte[32] ++ uint64 ++ byte[])`. The address values are serialized as a byte array of length 32 left-padded with zeroes, and all other value types are serialized according to the standard [transaction serialization](./tx_format.md#transaction). Note that the message data length is not included since there is only one dynamically sized field and can be implicitly determined by the hash preimage size.
+
+#### Output Message Nonce
+
+The nonce value for `OutputMessage` is computed as the [hash](./cryptographic_primitives.md#hashing) of the [Transaction ID](#transaction-id) that the message is an output for and the index of the output as a `uint8`. `hash(byte[32] ++ uint8)`
```
adlerjohn, line 55:
```suggestion
The nonce value for `OutputMessage` is computed as the [hash](./cryptographic_primitives.md#hashing) of the [Transaction ID](#transaction-id) that the message is an output for and the index of the output as a `uint8`: `hash(byte[32] ++ uint8)`.
```

adlerjohn (commented):
nit

specs/vm/instruction_set.md
adlerjohn: Ah this is already done!

adlerjohn (dismissed):

specs/vm/instruction_set.md
```diff
@@ -1416,6 +1417,53 @@ Cease VM execution and revert script effects. After a revert:
...
+- `$rA + $rB + 32` overflows
+- `$rA + $rB + 32 > VM_MAX_RAM`
+- `$rB > MEM_MAX_ACCESS_SIZE`
+- `$rB > MESSAGE_MAX_DATA_SIZE`
+- `$rC > tx.outputsCount`
+- In an external context, if `$rD > MEM[balanceOfStart(0), 8]`
+- In an internal context, if `$rD` is greater than the balance of asset ID 0 of output with contract ID `MEM[$fp, 32]`
+- `tx.outputs[$rC].type != OutputType.Message`
+- `tx.outputs[$rC].recipient != 0`
+- `MEM[$rA, 32] != 0`
```
vlopes11, line 1443:
```suggestion
- `MEM[$rA, 32] == 0`
```

I think we should assert the opposite here: panic only if the recipient is zero; not the opposite

vlopes11 (changes requested):

specs/vm/instruction_set.md
pixelcircuits: Right, good catch!

vlopes11 (approved):
LGTM

PR #326, opened 2022-06-06
Remove compressed tx format.

adlerjohn:
This isn't being used, and is a pain to maintain. Can be re-added when needed.


adlerjohn:
Waiting on @pixelcircuits to confirm

pixelcircuits (approved):

PR #338, opened 2022-06-06
Add dynamic jump instructions.

### LINKED ISSUE #325: Add dynamic jump instruction

adlerjohn:
Fixes #325

### DIFF
--- a/specs/vm/instruction_set.md
+++ b/specs/vm/instruction_set.md
@@ -36,8 +36,11 @@
 - [Control Flow Instructions](#control-flow-instructions)
   - [CIMV: Check input maturity verify](#cimv-check-input-maturity-verify)
   - [CTMV: Check transaction maturity verify](#ctmv-check-transaction-maturity-verify)
+  - [J: Jump](#j-jump)
   - [JI: Jump immediate](#ji-jump-immediate)
+  - [JNE: Jump if not equal](#jne-jump-if-not-equal)
   - [JNEI: Jump if not equal immediate](#jnei-jump-if-not-equal-immediate)
+  - [JNZ: Jump if not zero](#jnz-jump-if-not-zero)
   - [JNZI: Jump if not zero immediate](#jnzi-jump-if-not-zero-immediate)
   - [RET: Return from context](#ret-return-from-context)
 - [Memory Instructions](#memory-instructions)
@@ -702,6 +705,20 @@ Otherwise, advance the program counter `$pc` by `4`.
 
 See also: [BIP-65](https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki) and [Bitcoin's Time Locks](https://prestwi.ch/bitcoin-time-locks).
 
+### J: Jump
+
+|             |                                                     |
+|-------------|-----------------------------------------------------|
+| Description | Jumps to the code instruction offset by a register. |
+| Operation   | ```$pc = $is + $rA * 4;```                          |
+| Syntax      | `j $rA`                                             |
+| Encoding    | `0x00 rA - - -`                                     |
+| Notes       |                                                     |
+
+Panic if:
+
+- `$is + $rA * 4 > VM_MAX_RAM - 1`
+
 ### JI: Jump immediate
 
 |             |                                                |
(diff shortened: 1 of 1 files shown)


Voxelot (dismissed):

mohammadfawaz (approved):

otrho (commented):
Are we OK for the address register to be one of the reserved registers?  Should that be a panic if so?

adlerjohn:
> Are we OK for the address register to be one of the reserved registers?

I don't see why not. Maybe you want to jump to `$is` 🤷. It's at most a logic error, which is outside the purview of the VM.

otrho (approved):

specs/vm/instruction_set.md
```diff
@@ -702,6 +705,20 @@ Otherwise, advance the program counter `$pc` by `4`.
 
 See also: [BIP-65](https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki) and [Bitcoin's Time Locks](https://prestwi.ch/bitcoin-time-locks).
 
+### J: Jump
```
vlopes11, line 708: This is the only instruction with a single letter. What if we name it `JMP` or `JRE` (jump register)?

specs/vm/instruction_set.md
```diff
@@ -730,6 +761,20 @@ Panic if:
 
 - `$is + imm * 4 > VM_MAX_RAM - 1`
 
+### JNZ: Jump if not zero
```
vlopes11, line 764: Isn't `jnz $ra $rb` redundant to

```
jne $ra $zero $rb
```

specs/vm/instruction_set.md
```diff
@@ -716,6 +733,20 @@ Panic if:
 
 - `$is + imm * 4 > VM_MAX_RAM - 1`
 
+### JNE: Jump if not equal
```
vlopes11, line 736: Isn't `jne $ra $rb $rc` redundant to (or some optimized version of)

```
eq $rd $ra $rb
eq $re $rd $zero
mul $re $re $rc
add $rd $rd $re
j $rd
```

I understand we have `JNEI` and `JNZI` but I always saw them as special cases because of their immediate values. However, in this case, we have only runtime registers.

specs/vm/instruction_set.md
```diff
@@ -702,6 +705,20 @@ Otherwise, advance the program counter `$pc` by `4`.
 
 See also: [BIP-65](https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki) and [Bitcoin's Time Locks](https://prestwi.ch/bitcoin-time-locks).
 
+### J: Jump
+
+|             |                                                     |
+|-------------|-----------------------------------------------------|
+| Description | Jumps to the code instruction offset by a register. |
+| Operation   | ```$pc = $is + $rA * 4;```                          |
```
vlopes11, line 713: Should we concern with the case of `$rA == 0` that will cause an infinite loop and exhaust the gas or should we treat that as a malformed runtime execution? Note that this can't be trivially spotted at compilation time since it depends on runtime values.

Before, it was trivial to catch these cases since the jumps were immediate; therefore, part of the hardcoded program.

specs/vm/instruction_set.md
adlerjohn: I want to avoid `JR*` since we may want jump-relative in the future. `JMP` might not be bad.

specs/vm/instruction_set.md
adlerjohn: Ahhh you're right. The only reason that was needed was to free up 6 more bits of immediate. But without immediates, then it's redundant. Good catch!

specs/vm/instruction_set.md
adlerjohn: Haha you're not wrong, but conditional jumps will be most jumps. Definitely worth 1/5th the cost of them!

specs/vm/instruction_set.md
adlerjohn: It's not an issue since it's an application logic error, and DoS will be prevented by gas metering execution.

adlerjohn: Note also that the intended use of these instructions by a sane compiler is:
1. Load function pointer into register. Load current pc into a return register. Then jump to function pointer register.
2. When returning, jump to return register.

So the set of possible memory addresses is known at compile time.

specs/vm/instruction_set.md
adlerjohn: Fixed in [e5b39af](https://github.com/FuelLabs/fuel-specs/pull/338/commits/e5b39af7bcd5ec57fe6d7b63c2c70a4d7e0f3b00)

specs/vm/instruction_set.md
adlerjohn: Fixed in [94b4982](https://github.com/FuelLabs/fuel-specs/pull/338/commits/94b49821328ec2379fbdac202819cca4c97a405a)

vlopes11 (approved):
LGTM!

I think we should add some content somewhere explaining the rationale of allowing this approach in detriment of static analysis.

specs/vm/instruction_set.md
otrho: The actual `call` part will remain static most of the time, it's more about `ret`.  So (1) will most of the time just be 'save `pc` somewhere, jump to function'.

PR #341, opened 2022-06-09
Expanding the section on enum ABI encoding with more exmaples

### LINKED ISSUE #339: Expand the section on `enums` in the ABI spec a bit more
The [example provided](https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md#sum-types-enums) has two variants of the same size. There should be a more complex example that shows how to handled something like:
```rust
enum MySumType {
    X: b256, 
    y: u64,
}
```
Specifically, how to handle the padding if the chosen variant is `X`.

Also, the variants in the existing example should probably be capitalized.

I think there was agreement that the padding should be on the left, i.e. between the tag and the actual value:
Example:
```rust
enum E {
    X: b256,
    Y: u64,
}
```
Storing `E::Y(42)` into looks like:
```rust
word n:     0000000000000001  // This is the tag
word n+1:   0000000000000000  // First word of padding
word n+2:   0000000000000000  // Second word of padding
word n+3:   0000000000000000  // Third word of padding 
word n+4:   000000000000002a  // The actual data
```

mohammadfawaz:
Closes #339

Additional cases covered:
* Variants of different sizes
* Variants of type `()`.

Also capitalizing all variants in this file.

### DIFF
--- a/specs/protocol/abi.md
+++ b/specs/protocol/abi.md
@@ -401,8 +401,8 @@ foo(s(u64,bool))
 
 ```rust
 enum MyEnum {
-    foo: u64,
-    bar: bool,
+    Foo: u64,
+    Bar: bool,
 }
 fn foo(a: MyEnum);
 ```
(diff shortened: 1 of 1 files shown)


adlerjohn (approved):

digorithm (approved):
LGTM -- super clean 🙂

Issue #319, opened 2022-05-23
Document how arrays and tuples should be represented in the JSON ABI file and in the function selector (arrays only)


mohammadfawaz:
See https://github.com/FuelLabs/sway/pull/1632:

* Arrays in the JSON ABI file: insert their element types in the "components" section. Example:
If `[Person; 2]` is an argument to a contract method where `Person` is defined as `struct Person { name: str[4] }`, then the corresponding JSON ABI entry looks like:
```json
    {
        "name": "param1",
        "type": "[struct Person; 2]",
        "components": [
          {
            "name": "r#array_element",
            "type": "struct Person",
            "components": [
              {
                "name": "name",
                "type": "str[4]",
                "components": null
              }
            ]
          }
        ]
      },
```
* In the function selector, arrays are represented as `a[<type>;<size>]` in the strings that is used to generate the hash for the function selector (just as structs are represented as `s(...)`).

Edit: Removed the space after the semicolon in `a[<type>;<size>]`.

Edit 2: Tuples need to be handled correctly in the JSON ABI file (see https://github.com/FuelLabs/sway/pull/1653):
```json
    {
        "name": "param",
        "type": "(struct Person, u64)",
        "components": [
          {
            "name": "__tuple_element",
            "type": "struct Person",
            "components": [
              {
                "name": "age",
                "type": "u64",
                "components": null
              }
            ]
          },
          {
            "name": "__tuple_element",
            "type": "u64",
            "components": null
          }
        ]
      }
```

digorithm:
W.r.t `a[<type>; <size>]`, that's gonna be the actual string, right? As in, if I had `fn(a: [u64; 2])` the fn selector would be the first 4 bytes of the hash of the string `"[u64; 2]"`? Also, how about that space between `;` and `2`? Must have space or must _not_ have the space?

mohammadfawaz:
Very good point about the space. For consistency, I should probably take it out. I'll do that everywhere shortly.

And yeah, similarly to struct argument that show up as `s(..)`, I'm suggesting that arrays show up as `a[..]`.

digorithm:
Nice, so without the space, `"[u64;2]"`. 

And yes, `a[..]` makes sense to me!

mohammadfawaz:
Tuples were not handled properly as well. Fixing them in https://github.com/FuelLabs/sway/pull/1653 and I just updated the title and description of this issue accordingly.

PR #350, opened 2022-06-14
Clarify how the function selector is computed using a table and a more complex example

mohammadfawaz:
This is part 1 of https://github.com/FuelLabs/fuel-specs/issues/319.

Part 2 will handle the JSON ABI format.


digorithm (approved):
Nice!

PR #351, opened 2022-06-15
Encoding of each type in JSON ABI and other enhancements

mohammadfawaz:
Closes https://github.com/FuelLabs/fuel-specs/issues/319

Also:
* Some general enhancements
* Improved and more illustrative examples
* Removed the type `Byte`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ This repository specifies the Fuel protocol, including the Fuel Virtual Machine 
 1. [tx_validity.md](./specs/protocol/tx_validity.md): Defines transaction validity rules.
 1. [cryptographic_primitives](./specs/protocol/cryptographic_primitives.md): Cryptographic primitives used in Fuel.
 1. [abi.md](./specs/protocol/abi.md): ABI specifications.
+1. [storage_initialization.md](./specs/protocol/storage_initialization.md): JSON format for contract storage slot initialization.
 
 ### FuelVM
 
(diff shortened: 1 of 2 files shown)


specs/protocol/abi.md
```diff
@@ -6,92 +6,333 @@ This document describes and specifies the ABI (Application Binary Interface) of
...
+    Foo: u64,
+    Bar: bool,
+}
+struct MyStruct {
+    bim: u8,
+    bam: MyEnum,
+}
+
+abi MyContract {
+    fn complex_function(arg1: MyStruct, arg2: [b256; 4], arg3: (str[5], bool)) -> str[6];
```
adlerjohn, line 257: Can you make this example be an array of structs? It's not actually clear how that would look like.

specs/protocol/abi.md
```diff
@@ -137,11 +378,11 @@ _Important note:_ For the JSON representation of receipts, we represent 64-bit u
 
 ```json
 {
-    "type":"Panic",
-    "id":"0x39150017c9e38e5e280432d546fae345d6ce6d8fe4710162c2e3a95a6faff051",
-    "reason":"1",
-    "pc":"0xffffffffffffffff",
-    "is":"0xfffffffffffffffe"
+   "type":"Panic",
```
adlerjohn, line 381: Why go from 4 to 3 indentation?

specs/protocol/abi.md
mohammadfawaz: Done.

specs/protocol/abi.md
mohammadfawaz: So at first I thought `forc run --pretty-print` uses 3 spaces so I was running with 3 everywhere. Turns out that whatever JSON library we're using actually uses 2 spaces. Same goes for the JSON ABI file that the compiler generate. So, I'm now using 2 spaces everywhere.

adlerjohn (approved):

Issue #302, opened 2022-04-07
Specify format for initial storage values


sezna:
When a contract is deployed, we'd like to be able to specify the initial values held in state for certain slots. This will be dumped into the json abi file, most likely, and emitted by the compiler.

The values must be known at compile time, so the spec can be pretty low level. Like, we can just pass bytes straight to the slots.

mohammadfawaz:
What is the right approach here? Does adding a new array to the JSON file like this work?
```json
[
   {
      "type":"storage_initializer",
      "location":"0x0101010101010101010101010101010101010101010101010101010101010101",
      "initial_value":"0x1234123412341234123412341234123412341234123412341234123412341234",
   },
   {
      "type":"storage_initializer",
      "location":"0x0101010101010101010101010101010101010101010101010101010101010102",
      "initial_value":"0x1234123412341234123412341234123412341234123412341234123412340000",
   }
],
[
    
   {
      "type":"function",
      "inputs":[
         {
            "name":"arg",
            "type":"u64"
         }
      ],
      "name":"entry_one",
      "outputs":[

      ]
   }
]
```

Voxelot:
@mohammadfawaz this doesn't look like valid json to me, seems like we'd need to flatten these two arrays.

mohammadfawaz:
Yeah that makes sense. I had that first then I thought I'd split them into them into two arrays but I guess that's not valid json. Either way, does the rest make sense?

Voxelot:
While it is very clear and atomic, I'm wondering if this might be too verbose for contracts that initialize a lot of storage (e.g. structs with many fields etc). This could lead to very large JSON files that are slow to parse.

What if we came up with with a very simple binary encoding scheme for all initial storage, which uses only a single json field to describe the entire initial storage for a contract.

This could be something like:
```json
[
   {
      "type":"storage_initializer",
      "storage":"FLATTENED_INITIAL_STORAGE_BYTES",
   },
   {
      "type":"function",
      "inputs":[
         {
            "name":"arg",
            "type":"u64"
         }
      ],
      "name":"entry_one",
      "outputs":[

      ]
   }
]
```

Where `FLATTENED_INITIAL_STORAGE_BYTES` is simply concatenation of all storage keys and values.

You'd parse it by chunking like this:

```rust
const storage_slot_key_length = 32;
const storage_slot_value_length = 32;

let storage_slots = hex::decode(flattened_storage_bytes)
    .unwrap()
    .into_iter()
[...]

PR #352, opened 2022-06-15
Specify JSON format for storage initializers

mohammadfawaz:
Closes https://github.com/FuelLabs/fuel-specs/issues/302

~~Wasn't sure if the ABI file is the best location for this section. If not, I can move it to a different file.~~ Moved to a separate file.

### DIFF
--- a/specs/protocol/abi.md
+++ b/specs/protocol/abi.md
@@ -687,31 +687,3 @@ Calling `bar` with `MySumType::Z` yields:
 0x
 0000000000000002 // The discriminant of the chosen enum, in this case `2`.
 ```
-
-## JSON Format for Storage Initializers
-
-Contracts can request that certain storage slots are initialized to specific values. These initialized slots are represented in JSON format as an array where each element represents a storage slot and has the following properties:
-
-- `"key"`: String, a 32-byte key for a given storage slot;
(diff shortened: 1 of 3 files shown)


adlerjohn (commented):
I would put this in a different markdown file, since it'll be in a different file: this defines a JSON object, not a part of the JSON ABI.

specs/protocol/abi.md
```diff
@@ -687,3 +687,31 @@ Calling `bar` with `MySumType::Z` yields:
 0x
 0000000000000002 // The discriminant of the chosen enum, in this case `2`.
 ```
+
+## JSON Format for Storage Initializers
+
+Contracts can request that certain storage slots are initialized to specific values. These initialized slots are represented in JSON format as an array where each element represents a storage slot and has the following properties:
+
+- `"key"`: String, a 32-byte key for a given storage slot;
+- `"initial_value"`: String, a 32-byte value that initializes the slot;
```
adlerjohn, line 696: Since this is initializers, the fact that it's initial is implicit. Could shorter to just `value`.

adlerjohn (approved):
Looks fine to me.

Voxelot (approved):

PR #353, opened 2022-06-16
Add instructions to get tx fields

### LINKED ISSUE #287: Add instructions to get parts of tx

adlerjohn:
Fixes #287

Based on #318 since it's approved and should be merged in shortly.

Design decision: only two registers are used, one for input and out for output. That means it's not possible to get a nested variable-length field, e.g. if an input had a list of things (byte arrays notwithstanding). If that ever happens, this can be worked around in a backwards-compatible way by reading from the first register as well, similar to x86 instructions.

Implementation:

- [ ] fuel-asm
- [ ] fuel-vm

### DIFF
--- a/specs/vm/instruction_set.md
+++ b/specs/vm/instruction_set.md
@@ -1798,37 +1798,85 @@ Set `$rA` to `$fp->$fp` (i.e. `$rA` will point to the previous call frame's cont
 
 Get [fields from the transaction](../protocol/tx_format.md#transaction).
 
-| name                    | value   |
-|-------------------------|---------|
-| `GT_TYPE`               | `0x001` |
-| `GT_GAS_PRICE`          | `0x002` |
-| `GT_GAS_LIMIT`          | `0x003` |
-| `GT_MATURITY`           | `0x004` |
-| `GT_SCRIPT_LENGTH`      | `0x005` |
-| `GT_SCRIPT_DATA_LENGTH` | `0x006` |
-| `GT_INPUTS_COUNT`       | `0x007` |
-| `GT_OUTPUTS_COUNT`      | `0x008` |
-| `GT_WITNESSES_COUNT`    | `0x009` |
-| `GT_RECEIPTS_ROOT`      | `0x00A` |
-| `GT_SCRIPT`             | `0x00B` |
-| `GT_SCRIPT_DATA`        | `0x00C` |
-| `GT_INPUT_AT_INDEX`     | `0x00D` |
-| `GT_OUTPUT_AT_INDEX`    | `0x00E` |
-| `GT_WITNESS_AT_INDEX`   | `0x00F` |
-
-If `imm == GT_TYPE`:
+| name                                     | value   | set `$rA` to                                     |
+|------------------------------------------|---------|--------------------------------------------------|
+| `GT_TYPE`                                | `0x001` | `tx.type`                                        |
+| `GT_SCRIPT_GAS_PRICE`                    | `0x002` | `tx.gasPrice`                                    |
+| `GT_SCRIPT_GAS_LIMIT`                    | `0x003` | `tx.gasLimit`                                    |
+| `GT_SCRIPT_MATURITY`                     | `0x004` | `tx.maturity`                                    |
+| `GT_SCRIPT_SCRIPT_LENGTH`                | `0x005` | `tx.scriptLength`                                |
+| `GT_SCRIPT_SCRIPT_DATA_LENGTH`           | `0x006` | `tx.scriptDataLength`                            |
+| `GT_SCRIPT_INPUTS_COUNT`                 | `0x007` | `tx.inputsCount`                                 |
+| `GT_SCRIPT_OUTPUTS_COUNT`                | `0x008` | `tx.outputsCount`                                |
+| `GT_SCRIPT_WITNESSES_COUNT`              | `0x009` | `tx.witnessesCount`                              |
+| `GT_SCRIPT_RECEIPTS_ROOT`                | `0x00A` | Memory address of `tx.receiptsRoot`              |
+| `GT_SCRIPT_SCRIPT`                       | `0x00B` | Memory address of `tx.script`                    |
+| `GT_SCRIPT_SCRIPT_DATA`                  | `0x00C` | Memory address of `tx.scriptData`                |
+| `GT_SCRIPT_INPUT_AT_INDEX`               | `0x00D` | Memory address of `tx.inputs[$rB]`               |
+| `GT_SCRIPT_OUTPUT_AT_INDEX`              | `0x00E` | Memory address of `t.outputs[$rB]`               |
+| `GT_SCRIPT_WITNESS_AT_INDEX`             | `0x00F` | Memory address of `tx.witnesses[$rB]`            |
+| `GT_CREATE_GAS_PRICE`                    | `0x010` | `tx.gasPrice`                                    |
+| `GT_CREATE_GAS_LIMIT`                    | `0x011` | `tx.gasLimit`                                    |
+| `GT_CREATE_MATURITY`                     | `0x012` | `tx.maturity`                                    |
+| `GT_CREATE_BYTECODE_LENGTH`              | `0x013` | `tx.bytecodeLength`                              |
+| `GT_CREATE_BYTECODE_WITNESS_INDEX`       | `0x014` | `tx.bytecodeWitnessIndex`                        |
+| `GT_CREATE_STORAGE_SLOTS_COUNT`          | `0x015` | `tx.storageSlotsCount`                           |
+| `GT_CREATE_INPUTS_COUNT`                 | `0x016` | `tx.inputsCount`                                 |
+| `GT_CREATE_OUTPUTS_COUNT`                | `0x017` | `tx.outputsCount`                                |
+| `GT_CREATE_WITNESSES_COUNT`              | `0x018` | `tx.witnessesCount`                              |
+| `GT_CREATE_SALT`                         | `0x019` | Memory address of `tx.salt`                      |
+| `GT_CREATE_STORAGE_SLOT_AT_INDEX`        | `0x01A` | Memory address of `tx.storageSlots[$rB]`         |
+| `GT_CREATE_INPUT_AT_INDEX`               | `0x01B` | Memory address of `tx.inputs[$rB]`               |
+| `GT_CREATE_OUTPUT_AT_INDEX`              | `0x01C` | Memory address of `t.outputs[$rB]`               |
+| `GT_CREATE_WITNESS_AT_INDEX`             | `0x01D` | Memory address of `tx.witnesses[$rB]`            |
+| `GT_INPUT_TYPE`                          | `0x101` | `tx.inputs[$rB].type`                            |
+| `GT_INPUT_COIN_TX_ID`                    | `0x102` | Memory address of `tx.inputs[$rB].txID`          |
+| `GT_INPUT_COIN_OUTPUT_INDEX`             | `0x103` | `tx.inputs[$rB].outputIndex`                     |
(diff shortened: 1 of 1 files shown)


specs/vm/instruction_set.md
```diff
@@ -1785,3 +1785,50 @@ Panic if:
...
+| `GT_SCRIPT_DATA`        | `0x00C` |
+| `GT_INPUT_AT_INDEX`     | `0x00D` |
+| `GT_OUTPUT_AT_INDEX`    | `0x00E` |
+| `GT_WITNESS_AT_INDEX`   | `0x00F` |
+
+If `imm == GT_TYPE`:
+
+Panic if:
+
+- `$fp == 0` (in an external context)
```
Voxelot, line 1823: Looks like this is copy paste for gm which panics in external context when you pass an arg asking for contract metadata.
```suggestion
```

adlerjohn: Yeah it's not done 😂

otrho (commented):
This is a little confusing.
- It isn't obvious that the 'value' goes in the immediate arg.
- What should `$rB` be for fields which don't need an index?

adlerjohn:
Clarified in [9978d52](https://github.com/FuelLabs/fuel-specs/pull/353/commits/9978d521b802f81f16f5ec3b6d694719030a9292).

> It isn't obvious that the 'value' goes in the immediate arg.

Use the expression `imm` directly.

> What should `$rB` be for fields which don't need an index?

Explicitly specify that the value is ignored.

otrho (dismissed):
Maybe two instructions, one for indexed fields and one which doesn't need a second register might be nicer, no need to provide an `$rB` which is ignored, but having everything together in one place is probably better.

specs/vm/instruction_set.md
```diff
@@ -1785,3 +1786,101 @@ Panic if:
 - `$fp->$fp == 0` (if parent context is external)
 
 Set `$rA` to `$fp->$fp` (i.e. `$rA` will point to the previous call frame's contract ID).
+
+### GT: Get transaction fields
```
mohammadfawaz, line 1790: So we already have an instruction called `gt` (greater than). Should we call this one `gtf` instead?

otrho: 🤦 Whoops.

adlerjohn: Fixed in [4813ff0](https://github.com/FuelLabs/fuel-specs/pull/353/commits/4813ff0bc852b8af208b38d3ff791a7b60d4d1ff)

specs/vm/instruction_set.md
```diff
@@ -1785,3 +1786,101 @@ Panic if:
...
+| `GTF_SCRIPT_GAS_PRICE`                    | `0x002` | `tx.gasPrice`                                    |
+| `GTF_SCRIPT_GAS_LIMIT`                    | `0x003` | `tx.gasLimit`                                    |
+| `GTF_SCRIPT_MATURITY`                     | `0x004` | `tx.maturity`                                    |
+| `GTF_SCRIPT_SCRIPT_LENGTH`                | `0x005` | `tx.scriptLength`                                |
+| `GTF_SCRIPT_SCRIPT_DATA_LENGTH`           | `0x006` | `tx.scriptDataLength`                            |
+| `GTF_SCRIPT_INPUTS_COUNT`                 | `0x007` | `tx.inputsCount`                                 |
+| `GTF_SCRIPT_OUTPUTS_COUNT`                | `0x008` | `tx.outputsCount`                                |
+| `GTF_SCRIPT_WITNESSES_COUNT`              | `0x009` | `tx.witnessesCount`                              |
+| `GTF_SCRIPT_RECEIPTS_ROOT`                | `0x00A` | Memory address of `tx.receiptsRoot`              |
+| `GTF_SCRIPT_SCRIPT`                       | `0x00B` | Memory address of `tx.script`                    |
```
vlopes11, line 1814: Ditto; constant offset

vlopes11, line 1814: More relevant than the script offset, it might be worthy to add the functionality to fetch the script len

specs/vm/instruction_set.md
```diff
@@ -1785,3 +1786,101 @@ Panic if:
...
+| `GTF_SCRIPT_SCRIPT_DATA`                  | `0x00C` | Memory address of `tx.scriptData`                |
```
vlopes11, line 1815: Ditto; maybe provide script data len as well?

specs/vm/instruction_set.md
```diff
@@ -1785,3 +1786,101 @@ Panic if:
...
+| `GTF_SCRIPT_INPUT_AT_INDEX`               | `0x00D` | Memory address of `tx.inputs[$rB]`               |
```
vlopes11, line 1816: Ditto; len?

specs/vm/instruction_set.md
```diff
@@ -1785,3 +1786,101 @@ Panic if:
...
+| `GTF_SCRIPT_OUTPUT_AT_INDEX`              | `0x00E` | Memory address of `t.outputs[$rB]`               |
+| `GTF_SCRIPT_WITNESS_AT_INDEX`             | `0x00F` | Memory address of `tx.witnesses[$rB]`            |
```
vlopes11, line 1818: Ditto; len?

specs/vm/instruction_set.md
```diff
@@ -1785,3 +1786,101 @@ Panic if:
...
+| `GTF_TYPE`                                | `0x001` | `tx.type`                                        |
+| `GTF_SCRIPT_GAS_PRICE`                    | `0x002` | `tx.gasPrice`                                    |
+| `GTF_SCRIPT_GAS_LIMIT`                    | `0x003` | `tx.gasLimit`                                    |
+| `GTF_SCRIPT_MATURITY`                     | `0x004` | `tx.maturity`                                    |
+| `GTF_SCRIPT_SCRIPT_LENGTH`                | `0x005` | `tx.scriptLength`                                |
+| `GTF_SCRIPT_SCRIPT_DATA_LENGTH`           | `0x006` | `tx.scriptDataLength`                            |
+| `GTF_SCRIPT_INPUTS_COUNT`                 | `0x007` | `tx.inputsCount`                                 |
+| `GTF_SCRIPT_OUTPUTS_COUNT`                | `0x008` | `tx.outputsCount`                                |
+| `GTF_SCRIPT_WITNESSES_COUNT`              | `0x009` | `tx.witnessesCount`                              |
+| `GTF_SCRIPT_RECEIPTS_ROOT`                | `0x00A` | Memory address of `tx.receiptsRoot`              |
```
vlopes11, line 1813: This offset is constant. Is it worthy to add an operation to fetch it?

specs/vm/instruction_set.md
```diff
@@ -1785,3 +1786,101 @@ Panic if:
...
+| `GTF_WITNESS_DATA`                        | `0x302` | Memory address of `tx.witnesses[$rB].data`       |
+
+Panic if:
+
+- `$rA` is a [reserved register](./main.md#semantics)
+- `imm` is not one of the values listed above
+- The value of `$rB` results in an out of bounds access for variable-length fields
+- The input or output type does not match (`OutputChange` and `OutputVariable` count as `OutputCoin`)
+
+For fixed-length fields, the value of `$rB` is ignored.
```
vlopes11, line 1886: Isn't this redundant/too verbose for the specs?

adlerjohn: There's not really another case where a register's value is ignored, so better to be clear and unambiguous.

specs/vm/instruction_set.md
```diff
@@ -1785,3 +1786,101 @@ Panic if:
...
+| Operation   | Varies (see below).     |
+| Syntax      | `gtf $rA, $rB, imm`     |
+| Encoding    | `0x00 rA rB i i`        |
+| Notes       |                         |
+
+Get [fields from the transaction](../protocol/tx_format.md#transaction).
+
+| name                                      | `imm`   | set `$rA` to                                     |
+|-------------------------------------------|---------|--------------------------------------------------|
+| `GTF_TYPE`                                | `0x001` | `tx.type`                                        |
```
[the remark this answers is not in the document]
adlerjohn: That was removed in #311. This PR is based on the latest format (which isn't implemented yet), under the intention of the tx format being changed first then percolated up to the VM.

specs/vm/instruction_set.md
adlerjohn: That's provided with `GTF_SCRIPT_SCRIPT_DATA_LENGTH`

specs/vm/instruction_set.md
adlerjohn: The length of an input or output isn't really important since it's a struct instead of a list.

specs/vm/instruction_set.md
adlerjohn: I guess this wasn't the clearest, but the purpose of these instructions is to allow e.g. the stdlib to be able to get tx fields regardless of if additional fields are inserted and break the tx format. https://github.com/FuelLabs/fuel-specs/issues/287#issuecomment-1071890003

Without these instructions for example, we can't remove the byte price field without breaking the stdlib, which would be Very Bad for immutable smart contracts.

specs/vm/instruction_set.md
Voxelot, line 1804: Should we consider using imm values that are easily maskable based on their compatibility with the tx?

eg. all `GTF_SCRIPT_*` are compatible with a mask of `0x00ff`, all `GTF_CREATE_*` are under a mask of `0xff00`..

I noticed there is some degree of domain separation between the input, output and witness values, but there's not much room to grow if we add a new field for script transactions.

specs/vm/instruction_set.md
Voxelot, line 1818: Should we leave more spacing between script tx fields and create tx fields? I noticed you left plenty of space between the input, output and witness values, but script tx doesn't have much room to grow.

Also is there a way we could assign these imm values to make compatibility checking more efficient using bitmasks?

I.e. if script tx fields fit within a bitmask of `0x00ff` and create tx fields fit within a bitmask of `0xff00`, then it should be easier to specify which fields are not available. 

Right now this seems a little fuzzy by just saying:

> Panic If:
> The input or output type does not match (`OutputChange` and `OutputVariable` count as `OutputCoin`)

specs/vm/instruction_set.md
vlopes11: I see! Maybe add a tag to the compiler a flag on whether the user wants to compile for full gas optimization (so these opcodes are translated into hardcoded constants), or for long-term compatibility.

cc @mohammadfawaz @otrho @sezna

specs/vm/instruction_set.md
vlopes11: But aren't we creating a tech debt for when the user might want to iterate over all the inputs/outputs/witnesses?

adlerjohn: Can you clarify? If they want to iterate then they just iterate?

vlopes11: Lets say the user wants to iterate all the outputs of a transaction. Currently, he can fetch the output count in a constant offset of the canonical serialization.

Since we are shielding the user from long-term breaking changes in the structure of the VM, this deployed subroutine will no longer be valid when that offset changes because it will no longer be a constant he can fetch at runtime.

So, to set the number of iterations for the outputs, that can be trivially done - and breaking-changes safe - to fetch these from this function.

adlerjohn: Ah I think I see the confusion. You can get the number of inputs (i.e. the length of the list of inputs) with `GTF_SCRIPT_INPUTS_COUNT`. What you can't get is the _length in bytes_ of a particular input, because as above a single input is a struct, not a list.

vlopes11: Oh, overlooked that! Yes, we have these already implemented

PR #372, opened 2022-06-24
Check script and predicate / data lengths.

### LINKED ISSUE #357: Lengths must be checked to match
There are currently no requirements that lengths match the length/number of elements for transaction variable-length fields. They must match for the tx to be valid.

adlerjohn:
Fixes #357

Implementation:

- [ ] fuel-tx


Voxelot:
Since the contract bytecode root is computed the same way as scripts and predicates, should we perform the same validation on the create tx for contract bytecode?

adlerjohn:
It's already checked:

> `tx.data.witnesses[bytecodeWitnessIndex].dataLength != bytecodeLength * 4`

Voxelot (approved):

PR #377, opened 2022-06-28
Change the bullet points to headings in `/specs/protocol/abi.md` so that they can be clicked and used as direct links

### LINKED ISSUE #376: Change the bullet points to headings in `/specs/protocol/abi.md` so that they can be clicked and used as direct links
## Summary

The types are bullet pointed. 
If I want to reference a specific type then I must reference the parent heading and tell a user to scroll down to the relevant section. I should be able to directly link them to the correct section.

https://github.com/FuelLabs/fuel-specs/blob/e0d5d8aea3f6c344ca47776403accc58b67eeb01/specs/protocol/abi.md?plain=1#L78-L96

Braqzen:
Changes made

1. Each bullet point has been changed to be a h4 (####) heading.
2. Adjusted indentation

Closes #376

### DIFF
--- a/specs/protocol/abi.md
+++ b/specs/protocol/abi.md
@@ -75,167 +75,167 @@ the JSON representation of this ABI looks like:
 
 Below is a list of the JSON ABI formats for each available type:
 
-- `bool`:
+#### `bool`
 
-    ```json
+```json
+{
+  "name": "<var_name>",
+  "type": "bool",
+  "components": null
+} 
+```
+
+#### `u8`
+
+```json
+{
+  "name": "<var_name>",
+  "type": "u8",
+  "components": null
+} 
+```
+
+#### `u16`
+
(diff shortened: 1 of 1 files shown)


Braqzen:
Should I change the headings to not be wrapped in backticks so that they are not highlighted by the grey background?

mohammadfawaz:
> Should I change the headings to not be wrapped in backticks so that they are not highlighted by the grey background?

The grey background doesn't bother me personally. 

I really wanted this section to be a table like the table in https://github.com/FuelLabs/fuel-specs/blob/761b3c371929f21d48a9a215dfd27681db3dff15/specs/protocol/abi.md#function-selector-encoding, but I couldn't get multiline rows to look good :/

I think what you have here is good. I'll let John make the final call.

Thanks!

Issue #381, opened 2022-06-30
Introduce type arguments in JSON ABI and update the function selector with those arguments


mohammadfawaz:
Ref https://github.com/FuelLabs/sway/issues/1993

Potential solution: introduce a `typeArguments` field.

```rust
contract;

struct Foo<T, U> {
}

struct S {
    x: u64,
    y: b256,
}

abi MyContract {
    fn test_function(f: Foo<str[3], S>) -> bool;
}

impl MyContract for Contract {
    fn test_function(f: Foo<str[3], S>) -> bool {
        true
    }
}
```

JSON ABI:
```json
[
  {
    "type": "function",
    "inputs": [
      {
        "name": "f",
        "type": "struct Foo",
        "components": [],
        "typeArguments": [
          {
            "name": "T",
            "type": "str[3]",
            "components": null,
            "typeArguments": null
          },
          {
            "name": "U",
            "type": "struct S",
            "components": [
              {
                "name": "x",    
                "type": "u64",
                "components": null
                "typeArguments": null
              },
              {
                "name": "y",    
                "type": "b256",
                "components": null
                "typeArguments": null
              }
            ],
            "typeArguments": [] 
          }
        ]
      }
    ],
    "name": "test_function",
    "outputs": [
      {
        "name": "",
        "type": "bool",
        "components": null
        "typeArguments": null
      }
    ]
  }
]
```

PR #382, opened 2022-07-01
Type arguments in the JSON ABI and the function selector

mohammadfawaz:
Closes https://github.com/FuelLabs/fuel-specs/issues/381

### DIFF
--- a/specs/protocol/abi.md
+++ b/specs/protocol/abi.md
@@ -316,7 +316,7 @@ enum MyEnum<V> {
     Bar: bool,
 }
 struct MyStruct<T, U> {
-    bim: u8,
+    bim: T,
     bam: MyEnum<u64>,
 }
 struct MyOtherStruct {
@@ -347,8 +347,15 @@ its JSON representation would look like:
         "components": [
           {
             "name": "bim",
-            "type": "u8",
-            "components": null,
+            "type": "[b256; 3]",
+            "components": [
+              {
+                "name": "__array_element",
+                "type": "b256",
+                "components": null,
+                "typeArguments": null
+              }
+            ],
             "typeArguments": null
           },
           {
@@ -410,7 +417,7 @@ its JSON representation would look like:
             "components": [
               {
                 "name": "bim",
-                "type": "u8",
+                "type": "u64",
                 "components": null,
                 "typeArguments": null
               },
@@ -835,7 +842,7 @@ enum MyEnum<V> {
     Bar: bool,
 }
 struct MyStruct<T, U> {
-    bim: u8,
+    bim: T,
     bam: MyEnum<u64>,
 }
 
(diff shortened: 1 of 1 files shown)


vnepveu (dismissed):
EDIT: I have questions

vnepveu (commented):
Sorry for all the edits,  completely misunderstood at first lol. I think it's clear enough! Maybe the example could make use of the generic type. Also

specs/protocol/abi.md
```diff
@@ -244,17 +311,25 @@ Below is a list of the JSON ABI formats for each available type:
 Given the following ABI declaration:
 
 ```rust
-enum MyEnum {
+enum MyEnum<V> {
     Foo: u64,
     Bar: bool,
 }
-struct MyStruct {
+struct MyStruct<T, U> {
```
vnepveu, line 318: Maybe the `U` and `V` types could be used by the components themselves? ie have `Foo: V`. This would make the example more clear.

mohammadfawaz: I think you're right. I wanted to emphasize that the generic types don't really need to be used by the data memebers (e.g. Vec<T>`). However, maybe a good compromise is to use one generic type and not use the other.

vnepveu: Yes I think that's a good idea!

mohammadfawaz: Done i n https://github.com/FuelLabs/fuel-specs/pull/382/commits/fdf55f4b6b0e3073a1da5f8ea5b28cd8f272d768

vnepveu (approved):
LGTM!

digorithm (approved):
Nice -- LGTM!

When's the compiler planning on delivering this? Asking because it needs a _lot_ of work on the SDKs as well.

mohammadfawaz:
> Nice -- LGTM!
> 
> When's the compiler planning on delivering this? Asking because it needs a _lot_ of work on the SDKs as well.

I have the change ready but it needs [this type](https://github.com/FuelLabs/fuels-rs/blob/43a43964e7caec632c360a0987d3ef34b7fc10bc/packages/fuels-types/src/lib.rs#L29) to be updated. I'm thinking I'll workaround that using a temporary type in `sway-types` until it's updated in `fuels-rs`.

mohammadfawaz:
@adlerjohn is this okay to be merged? Just want the green light from you as the owner of the specs 😄

specs/protocol/abi.md
```diff
@@ -332,22 +460,38 @@ its JSON representation would look like:
...
+        "type": "struct MyOtherStruct",
+        "components": [
+          {
+            "name": "bom",
+            "type": "u64",
+            "components": null,
+            "typeArguments": null
+          }
+        ],
+        "typeArguments": []
```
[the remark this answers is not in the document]
mohammadfawaz: If the struct/enum does not have generic parameters, `typeArguments` is an empty array. We can make that `null` if we want.

Most `typeArguments` in this example are indeed `null` but there are some that are not (e.g. line 388)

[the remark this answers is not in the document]
adlerjohn: Ah I see now on 388. How about this one? Shouldn't there never by an empty-array for type argument?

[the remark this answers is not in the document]
mohammadfawaz: Changed this one to `null` and adjusted the wording. This is probably more consistent.

adlerjohn (approved):

PR #383, opened 2022-07-01
Formalize Double-spend Validation

Voxelot:
We already implemented these checks, but they are missing from the specs.

### DIFF
--- a/specs/protocol/tx_format.md
+++ b/specs/protocol/tx_format.md
@@ -62,9 +62,12 @@ Transaction is invalid if:
 - `inputsCount > MAX_INPUTS`
 - `outputsCount > MAX_OUTPUTS`
 - `witnessesCount > MAX_WITNESSES`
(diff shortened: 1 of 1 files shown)


pixelcircuits:
Do we also need to specify the validity of the amount on `InputCoin` and amount/sender/recipient/data on `InputMessage`?

Voxelot:
Most likely, although that would be separate from tx format validation and requires interaction with state.

pixelcircuits (approved):

adlerjohn (approved):

vlopes11 (approved):
LGTM

As a nit, it might be worthy to use the same terminology both in the code and in the specs.

In the code, for instance, it is named as "asset id" and "utxo id".

https://github.com/FuelLabs/fuel-tx/blob/aacd5624a98242f474c1989b131a3a4771b61ba4/src/transaction/validation.rs#L252-L260

PR #384, opened 2022-07-02
fix eager rounding, including missing byte costs in resultant fee

Voxelot:
> gasBalance = ceiling((gasPrice * gasLimit) / GAS_PRICE_FACTOR)
> bytesBalance = ceiling((size(tx) * gasPrice * GAS_PER_BYTE) / GAS_PRICE_FACTOR)
> return sentBalance + gasBalance + bytesBalance

- When calculating unavailableBalance, we shouldn't round gas separately from bytes as the minimum unavailableBalance will be `2` instead of `1`, even though we would only charge an amount of `1`. @luizstacio raised this to my attention.

> The fees incurred for a transaction are `ceiling(((tx.gasLimit - unspentGas) * tx.gasPrice) / GAS_PRICE_FACTOR)`.
- Bytes were not included in the final post-execution fee calculation. If `tx.gasLimit` and `tx.gasPrice` are intended to be the sole determining factors in how much a tx costs, we'd need to make the following changes instead:
  - exclude byte calculations from the unavailableBalance
  - initialize G_GAS and C_GAS with `tx.gasLimit - (size(tx) * GAS_PER_BYTE)` instead of `tx.gasLimit`

### DIFF
--- a/specs/protocol/tx_validity.md
+++ b/specs/protocol/tx_validity.md
@@ -107,13 +107,15 @@ def unavailable_balance(tx, col) -> int:
     monotonic and the cost of bytes should approximately makes up for this.
     """
     sentBalance = sum_outputs(tx, col)
-    gasBalance = ceiling((gasPrice * gasLimit) / GAS_PRICE_FACTOR)
+    gasBalance = gasPrice * gasLimit / GAS_PRICE_FACTOR
     # Size excludes witness data as it is malleable (even by third parties!)
-    bytesBalance = ceiling((size(tx) * gasPrice * GAS_PER_BYTE) / GAS_PRICE_FACTOR)
+    bytesBalance = size(tx) * GAS_PER_BYTE * gasPrice / GAS_PRICE_FACTOR
+    # Total fee balance
+    feeBalance = ceiling(gasBalance + bytesBalance)
     # Only base asset can be used to pay for gas
     if col != 0:
         return sentBalance
-    return sentBalance + gasBalance + bytesBalance
+    return sentBalance + feeBalance
 
 return available_balance(tx, col) >= unavailable_balance(tx, col)
 ```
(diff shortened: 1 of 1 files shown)


adlerjohn (commented):
merge conflicts

adlerjohn (commented):
So, functionally the change is fine if we're starting with the axiom that gas price should be the sole way to scale price per resource. I actually don't think that should be the case, as in the future we'll almost certainly have multi-dimensional resource pricing.

While I do see that the proposed change is a pragmatic simplification for the immediate reality, my worry is that unifying resources like this might result in developers making incorrect assumptions around the invariants of the protocol.

That being said, I'm not 100% opposed to the proposed change, and could be convinced that it'll be fine and everything will work out.

Voxelot:
If the amount we actually end up charging users is the sum total rounded, vs price components rounded and then added, then the calculation of the unavailable balance should have the same behavior.

For example, let's assume there are multiple price dimensions and each has its own respective max unit limit. The amount charged would still be `ceiling( sum of all fee dimensions based on usage )`, regardless of the gas price factor. When computing the unavailable balance, this should also be `ceiling( sum of all potential fees based on limits )`, rather than each component being rounded before adding.

adlerjohn:
> then the calculation of the unavailable balance should have the same behavior

Yes, agree with that much.

> Actually, it doesn't make sense to round all values together unless they share the same price_factor divisor

Yeah, that was what I was getting at: these are two resources that are fundamentally orthogonal and the fact that they currently have the same scaling is an artifact rather than an intentional design decision.

> I still think we should only split out the rounding once we actually have different gas_price_factors, otherwise we end up reserving more unavailable balance than needed.

Fair enough.

adlerjohn (approved):

PR #388, opened 2022-07-10
fix: fix argument name in example

vnepveu:


mohammadfawaz (approved):
Thanks for catching this!

PR #389, opened 2022-07-11
Docs: Add SMT test spec and test fixtures

bvrooman:
Related issues: 
 - Progress on https://github.com/FuelLabs/fuel-merkle/issues/80

This PR adds the SMT test specification and related test fixtures currently living in the `fuel-merkle` repo to the `fuel-specs` repo. The `fuel-specs` repo is the appropriate housing for these documents since they describe universally applicable test outputs and are agnostic to any implementation. Sparse Merkle tree implementations can adhere to the SMT test specifications in `fuel-specs` to ensure correctness; this includes the `fuel-merkle` crate and `fuel-merkle-sol` library. 

SMT libraries can include the `fuel-specs` repo as a git submodule in order to pull in the test fixtures.

In a future PR, I will remove the existing test spec and fixture files from `fuel-merkle` and specify `fuel-specs` as a submodule.


Voxelot:
would it be more accurate to call these `test_vectors` instead of `fixtures`? Fixtures are typically just the environment / setup ingredients for a test rather than the actual SUT.

bvrooman:
Yea, I think that's a good distinction. I've updated the folder name to `test_vectors`.

Voxelot:
Git submodules can be a little tricky to manage sometimes.. Is there a way we could use git hooks to automatically pull these vectors into fuel-merkle?

bvrooman:
I agree that submodules can be tricky: codebases that uses submodules have to remember to run all the submodule commands, i.e. `git submodule init`, `git submodule update`, etc, and clients have to stay on top of updates to the source repo manually.

I can look into alternative solutions, like git hooks. If can automate pulling in the test vectors, that would be great.

That said, can we get the test vectors into `fuel-specs` in the meantime? I believe that this step will still need to come first regardless of whichever solution we go with.

adlerjohn:
Verification is easier than execution after all, so my preference would instead be a CI workflow in `fuel-merkle` that pulls the test vectors from the latest `master` of `fuel-specs` and compares for diffs.

adlerjohn (approved):
utACK

PR #391, opened 2022-07-12
Fix the JSON ABI of the complex example

mohammadfawaz:
The type arguments for `MyStruct` in `"[struct MyStruct; 4]"` were missing.

Corresponding change in the compiler: https://github.com/FuelLabs/sway/pull/2303


vnepveu (approved):
nice

PR #394, opened 2022-08-01
Set contract input values in validity rules instead of tx format

Voxelot:
We had somewhat conflicting definitions of when and how contract input values should be set between tx_format and tx_validity. This PR moves the description of how to set the contract inputs all to the validity rules.

### DIFF
--- a/specs/protocol/tx_format.md
+++ b/specs/protocol/tx_format.md
@@ -244,7 +244,7 @@ Note: when signing a transaction, `txID`, `outputIndex`, `balanceRoot`, `stateRo
 
 Note: when verifying a predicate, `txID`, `outputIndex`, `balanceRoot`, `stateRoot`, and `txPointer` are initialized to zero.
 
-Note: when executing a script, `txID`, `outputIndex`, `balanceRoot`, `stateRoot` and `txPointer`  are initialized to the transaction ID, output index, amount, state root and txPointer of the contract with ID `contractID`.
+Note: when executing a script, `txID`, `outputIndex`, `balanceRoot`, `stateRoot` and `txPointer` are initialized to the transaction ID, output index, amount, state root and txPointer of the contract with ID `contractID`.
 
 ### InputMessage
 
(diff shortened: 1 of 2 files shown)


vlopes11 (approved):

adlerjohn (changes requested):
Wait no, you need to initialize contract inputs with the correct roots since the roots are updated in-place during execution of the script. It's needed for fraud proofs.

Voxelot:
> Wait no, you need to initialize contract inputs with the correct roots since the roots are updated in-place during execution of the script. It's needed for fraud proofs.

It's not clear to me what the difference is between setting these fields vs the txpointer. Seems like they would all be set at the same time?

specs/protocol/tx_validity.md
```diff
@@ -76,7 +76,7 @@ for input in tx.inputs:
 return True
 ```
 
-If this check passes, the UTXO ID `(txID, outputIndex)` fields of each contract input is set to the UTXO ID of the respective contract. The `txPointer` of each input is also set to the TX pointer of the UTXO with ID `utxoID`.
+If this check passes, the UTXO ID `(txID, outputIndex)`, `stateRoot`, and `balanceRoot` fields of each contract input are set to the current values of the respective contract. The `txPointer` of each input is also set to the TX pointer of the UTXO with ID `utxoID`.
```
adlerjohn, line 79: This change seems fine, but the last part should be reworded to exclude message inputs.

specs/protocol/tx_format.md
```diff
@@ -244,7 +244,7 @@ Note: when signing a transaction, `txID`, `outputIndex`, `balanceRoot`, `stateRo
 
 Note: when verifying a predicate, `txID`, `outputIndex`, `balanceRoot`, `stateRoot`, and `txPointer` are initialized to zero.
 
-Note: when executing a script, `txID`, `outputIndex`, `balanceRoot`, and `stateRoot` are initialized to the transaction ID, output index, amount, and state root of the contract with ID `contractID`, and `txPointer` is initialized to zero.
+Note: when executing a script, `txID`, `outputIndex`, `balanceRoot`, `stateRoot` and `txPointer` are initialized to zero.
```
adlerjohn, line 247: Rather than initializing everything to zero, which is bad since we need the state and balance root set properly so they can be updated during execution for fraud proofs...you should instead set the `txPointer` properly! I.e. 804f6797a33e3727080c03b2d7c3d2ccf5dc92d3 was fine.

xgreenx:
@Voxelot Seems we can close it?=)

PR #396, opened 2022-08-03
Add timestamp instruction.

### LINKED ISSUE #220: Add Timestamp Opcode
# Motivation
Currently, the FuelVM spec does not include a timestamp. This is for many practical reasons, namely, that we do not know when the block will be produced precisely and thus a timestamp could be manipulated quite heavily.

# Solution
I propose a timestamp opcode `timestamp` which is enforced by by using both the underlying layer-1 chain and by checking the previous blocks stamp.

The timestamp for each Fuel block must be both within 30 minutes of the underlying related Ethereum block (included in the header), and greater than the previous Fuel blocks posted timestamp.

# Note
The timestamp should not be relied upon for highly precision oriented use-cases which need more than 30 minutes of precision. But for many general uses cases (i.e. vesting schedules etc), this would work quite well.

# Opcode
I expect the opcode to only be a single op with no arguments. Similar to the block number op.

adlerjohn:
Closes #220

Two considerations:

1. Do we want to get the timestamp of a given block, instead of the current block? Would require an addition instruction (block height) before using this if you wanted to get the time of the current block, but would open up the door to querying past block timestamps.
2. What time format do we want? I suggest TAI64 since it's 64-bit and there's a Rust crate for it.

Implementation:

- [x] fuel-asm: https://github.com/FuelLabs/fuel-asm/issues/88
- [ ] fuel-vm: https://github.com/FuelLabs/fuel-vm/issues/183

### DIFF
--- a/specs/vm/instruction_set.md
+++ b/specs/vm/instruction_set.md
@@ -74,7 +74,7 @@
   - [SRWQ: State read 32 bytes](#srwq-state-read-32-bytes)
   - [SWW: State write word](#sww-state-write-word)
   - [SWWQ: State write 32 bytes](#swwq-state-write-32-bytes)
-  - [TIME: Current timstamp](#time-current-timstamp)
+  - [TIME: Timstamp at height](#time-timstamp-at-height)
   - [TR: Transfer coins to contract](#tr-transfer-coins-to-contract)
   - [TRO: Transfer coins to output](#tro-transfer-coins-to-output)
 - [Cryptographic Instructions](#cryptographic-instructions)
@@ -1489,7 +1489,7 @@ Panic if:
 - `$rB + 32 > VM_MAX_RAM`
 - `$fp == 0` (in the script context)
 
-### TIME: Current timstamp
+### TIME: Timstamp at height
 
 |             |                                         |
 |-------------|-----------------------------------------|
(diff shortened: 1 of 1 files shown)


vlopes11:
1. How about we take a register (with the block height), and an immediate value that, if is set as `1`, will ignore the register and take the current block height? This way, we cover all scenarios
2. `tai64` is a reliable standard and I think it is a solid call. However, the rust implementation seems to be quite questionable as it is just wrapping around [SystemTime::duration_since](https://doc.rust-lang.org/std/time/struct.SystemTime.html#method.duration_since), when the actual monotonic clock is [Instant](https://doc.rust-lang.org/std/time/struct.Instant.html).

https://github.com/RustCrypto/formats/blob/b289ef6a40bd9c8927d78c4e8dabe3af78a4aa6d/tai64/src/lib.rs#L187-L193

I'm not saying it is wrong since I just fast-forward the lines of code, but it might need some investigation. Worst case scenario, we can create a binding to `libtai`, that is known to be safe and sound. Anyway, that would just be an implementation detail.

adlerjohn:
> How about we take a register (with the block height), and an immediate value that, if is set as 1, will ignore the register and take the current block height? This way, we cover all scenarios

That's actually not a bad idea!

As for the actual time format, looks like there's no objection to using 64-bits at the granularity of seconds. Beyond that yeah it's more of a detail we can bikeshed later.

adlerjohn:
Ready for re-review @vlopes11

vlopes11 (approved):

Voxelot (approved):

PR #400, opened 2022-08-14
Removed owner from input messages

pixelcircuits:
This PR moves what previously was referred to as the "owner" from input messages into the "recipient" field. This removes redundancy for messages that didn't need a "recipient" field (like a simple ETH bridging). What was previously used in the "recipient" field can simply be placed in the "data" field. This also simplifies messages and makes the calculation of messageId the same for both input and output messages.

### DIFF
--- a/specs/protocol/abi.md
+++ b/specs/protocol/abi.md
@@ -767,11 +767,11 @@ _Important note:_ For the JSON representation of receipts, we represent 64-bit u
 #### MessageOut receipt
 
 - `type`: `MessageOut`.
-- `messageID`: Hexadecimal string representation of the 256-bit (32-byte) message ID as described [here](../protocol/identifiers.md#output-message-id).
+- `messageID`: Hexadecimal string representation of the 256-bit (32-byte) message ID as described [here](../protocol/identifiers.md#message-id).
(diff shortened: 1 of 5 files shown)


adlerjohn (approved):
Looks good to me! Would like a second set of eyes from @Voxelot or @vlopes11 to confirm.

Voxelot:
Would it make sense to use owner instead of recipient? That way it's more consistent with our standard UTXO's which use the owner field for specifying TX output recipients.

pixelcircuits:
> Would it make sense to use owner instead of recipient? That way it's more consistent with our standard UTXO's which use the owner field for specifying TX output recipients.

I left the term "recipient" because it's currently describing all messages (both input and output). "Owner" doesn't make sense for output messages because it will only ever be addresses. It's kind of it's own separate thing

Voxelot (approved):
> I left the term "recipient" because it's currently describing all messages (both input and output). "Owner" doesn't make sense for output messages because it will only ever be addresses. It's kind of it's own separate thing

"Recipient" does make sense in the context of "sending and receiving messages". The difference between owner and recipient isn't a big deal and probably not worth bikeshedding.

PR #405, opened 2022-09-14
Specify block header

### LINKED ISSUE #284: Block header spec
Fields:
1. MMR of previous block header hashes (binary Merkle tree is an MMR)
2. Coinbase address
3. Fuel block height
4. Ethereum block number (!) (current, not finalized)
5. timestamp (!) (current)
6. Transactions root (binary Merkle tree, not MST)
7. number of txs
8. next validator set root
9. next voting power
10. Hash of Tendermint commit for previous block (>2/3 of voting power) (canonical commit, client needs to broadcast this _with_ proposal)
11. root of withdrawal IDs

Separate from block header:
1. Tendermint Commit for current block (verified by the contract)

Note: consider wrapping core, consensus, and Ethereum specific fields

adlerjohn:
Fixes #284

Adds a specification for the block header. The header is split into two: the application header (i.e. the Fuel transactions) and the consensus header (i.e. the blockchain + consensus logic), which wraps the application header. While not strictly necessary, this allows for a cleaner separation of concerns logically. From an implementation perspective, this separation doesn't actually need to be done today.

The intention is that a consensus signature for the current block will be sent together with that block (header) to the base layer as a means of authentication. This allows an untrusted relayer to relay signed block headers to the base chain.


specs/protocol/block_header.md
```diff
@@ -0,0 +1,24 @@
...
+## Application Header
+
+name                 | type       | description
+---------------------|------------|-----------------------------------------------------------------------------------------------------
+`height`             | `uin64`    | Height of this block.
+`timestamp`          | `uin64`    | Time this block was created, in [TAI64](https://cr.yp.to/libtai/tai64.html) format.
+`txCount`            | `uint64`   | Number of [transaction](./tx_format.md#transaction)s in this block.
+`txRoot`             | `byte[32]` | [Merkle root](./cryptographic_primitives.md#binary-merkle-tree) of [transaction](./tx_format.md#transaction)s in this block.
+`outputMessagesCount`|`uint64`| Number of [output message](./tx_format.md#outputmessage)s in this block.
+`outputMessagesRoot` | `byte[32]` | [Merkle root](./cryptographic_primitives.md#binary-merkle-tree) of [output message](./tx_format.md#outputmessage)s in this block.
```
adlerjohn, line 12: Do we need messages as separate from the transactions here? @pixelcircuits

Voxelot: seems like they should be derivable from the transaction root?

adlerjohn: It is, yes. But if validators don't sign over it then there's no compact way of posting such a root in an authenticated way to an Ethereum contract.

pixelcircuits: My thinking was for validators to sign over the entire state of output messages set per block, that way we only have to periodically commit to Ethereum rather than commit a hash/root for each and every block.

Voxelot: > It is, yes. But if validators don't sign over it then there's no compact way of posting such a root in an authenticated way to an Ethereum contract.

Wouldn't this apply to all the data posted to ethereum, and not just message outputs? How does splitting messages out into a separate field address this issue?

> My thinking was for validators to sign over the entire state of output messages set per block, that way we only have to periodically commit to Ethereum rather than commit a hash/root for each and every block.

Let's assume the validators signed the current block hash MMR root, and the commit also includes a proof that the current block is linked to the last committed block. Seems like that may be sufficient to prove any of the other fields in previous blocks that may or may not have been committed to eth? 

This would also require a commit every time the validator staking set is modified, but that could be amortized using epochs. I.e. staking adjustments are batched up and only applied every epoch (some arbitrary duration like ~1000 blocks).

specs/protocol/block_header.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Consensus Header
+
+name                   | type       | description
+-----------------------|------------|-----------------------------------------------------------------------------------------------------------------------
+`prevRoot`           | `byte[32]` | [Merkle root](./cryptographic_primitives.md#binary-merkle-tree) of all previous [consensus header](#consensus-header) hashes (i.e. not including this block).
+`applicationHash`      | `byte[32]` | [Hash](./cryptographic_primitives.md#hashing) of serialized [application header](#application-header).
+`nextValidatorSetRoot` | `byte[32]` | [Merkle root](./cryptographic_primitives.md#binary-merkle-tree) of the validator set that will vote on the next block.
+`nextVotingPower`      | `uint64`   | Voting power of the validator set that will vote on the next block.
+`prevCommitHash`       | `byte[32]` | [Hash](./cryptographic_primitives.md#hashing) of consensus commit for `prevRoot`.
```
vlopes11, line 22: Is it overkill to make votes merkelized so we can efficiently prove that a given vote is in the set of votes that defined this block?

specs/protocol/block_header.md
pixelcircuits: We _could_ just have the proof travel down the chain to a certain block first. I'm not crazy enthusiastic about this idea though, since the proof size would grow almost linearly depending on when the message was outputted vs the next commit to Ethereum. I think this would end up in users trying to time txs to right before the state gets committed which could cause regular periodic spikes in L2 gas fees.

pixelcircuits: How about for now, we remove the output messages from being their own thing in the "header" as they can be proven/derived from the tx root. Then when we figure out the concept of "epochs" (which will be needed to freeze the validator set to prevent having to check in every single block to Ethereum) we have it include committing a rolled up merkle tree of all message outputs in the past epoch

specs/protocol/block_header.md
adlerjohn: It's not overkill, but there's also not really any usecase for that. The contract verifies all votes as they come in, so anything after that should just be as cheap as possible.

specs/protocol/block_header.md
adlerjohn: > How about for now, we remove the output messages from being their own thing in the "header" as they can be proven/derived from the tx root.

That would require users that want to consume an output message to provide an entire transaction to Ethereum to be parsed. If the output messages are in their own tree, then only the output message itself needs to be proven.

specs/protocol/block_header.md
```diff
@@ -0,0 +1,24 @@
...
+`outputMessagesCount`|`uint64`| Number of [output message](./tx_format.md#outputmessage)s in this block.
+`outputMessagesRoot` | `byte[32]` | [Merkle root](./cryptographic_primitives.md#binary-merkle-tree) of [output message](./tx_format.md#outputmessage)s in this block.
+
+## Consensus Header
+
+name                   | type       | description
+-----------------------|------------|-----------------------------------------------------------------------------------------------------------------------
+`prevRoot`           | `byte[32]` | [Merkle root](./cryptographic_primitives.md#binary-merkle-tree) of all previous [consensus header](#consensus-header) hashes (i.e. not including this block).
+`applicationHash`      | `byte[32]` | [Hash](./cryptographic_primitives.md#hashing) of serialized [application header](#application-header).
+`nextValidatorSetRoot` | `byte[32]` | [Merkle root](./cryptographic_primitives.md#binary-merkle-tree) of the validator set that will vote on the next block.
```
adlerjohn, line 20: Note: consider epochs here

specs/protocol/block_header.md
Voxelot: Having the message merkle root separated from txroot would help to limit the amount of leaf data that needs to be proved by users. Since there's not really much use case to prove anything else in a tx currently.

adlerjohn:
Converting to draft to separate configurations (PoA, PoS, etc.).

specs/protocol/block_header.md
```diff
@@ -0,0 +1,24 @@
...
+name                 | type       | description
+---------------------|------------|-----------------------------------------------------------------------------------------------------
+`height`             | `uin64`    | Height of this block.
+`timestamp`          | `uin64`    | Time this block was created, in [TAI64](https://cr.yp.to/libtai/tai64.html) format.
+`txCount`            | `uint64`   | Number of [transaction](./tx_format.md#transaction)s in this block.
+`txRoot`             | `byte[32]` | [Merkle root](./cryptographic_primitives.md#binary-merkle-tree) of [transaction](./tx_format.md#transaction)s in this block.
+`outputMessagesCount`|`uint64`| Number of [output message](./tx_format.md#outputmessage)s in this block.
+`outputMessagesRoot` | `byte[32]` | [Merkle root](./cryptographic_primitives.md#binary-merkle-tree) of [output message](./tx_format.md#outputmessage)s in this block.
+
+## Consensus Header
```
Voxelot, line 14: We should have different consensus headers based on the network configuration. For example, we won't need things like validatorSetRoot or votingPower for PoA, and a PoA sidechain contract will likely be a lot simpler as a consequence.

Voxelot: It would also be good if the consensus header included enough info to be authenticated by the contract, rather than only allowing the PoA node to post blocks. This way anyone could post block headers to Eth.

specs/protocol/block_header.md
```diff
@@ -0,0 +1,24 @@
+# Block Header
+
+## Application Header
+
+name                 | type       | description
+---------------------|------------|-----------------------------------------------------------------------------------------------------
+`height`             | `uin64`    | Height of this block.
+`timestamp`          | `uin64`    | Time this block was created, in [TAI64](https://cr.yp.to/libtai/tai64.html) format.
+`txCount`            | `uint64`   | Number of [transaction](./tx_format.md#transaction)s in this block.
+`txRoot`             | `byte[32]` | [Merkle root](./cryptographic_primitives.md#binary-merkle-tree) of [transaction](./tx_format.md#transaction)s in this block.
```
pixelcircuits, line 10: can we move the `outputMessageCount` to be right after the `txRoot` so we have 4 uint64s in a row? This will help reduce data in the solidity implementation

pixelcircuits:
What is the usefulness of `prevCommitHash`? What would want it?

adlerjohn:
> What is the usefulness of `prevCommitHash`? What would want it?

In PoS mode (i.e. not actually relevant for this PR and will be removed) it serves as a canonical commitment to the consensus commit (a poor name for a set of votes). Without that, consensus votes aren't actually committed to anywhere.

specs/protocol/block_header.md
adlerjohn: Fixed in [6a91e0b](https://github.com/FuelLabs/fuel-specs/pull/405/commits/6a91e0b01b04db45f2c0a36c1d061c73e37cdbf4)

specs/protocol/block_header.md
adlerjohn: Fixed in [b395226](https://github.com/FuelLabs/fuel-specs/pull/405/commits/b395226ce46f3f1e1a5a431e58a6eea661e2dd25)

pixelcircuits:
Can we move the 'height' field into the consensus header? This would allow the commit logic on Ethereum to only need the consensus header for validation. It's also nice because it tells us how big the 'prevRoot' tree is.

adlerjohn:
@pixelcircuits moved height and timestamp into the consensus header. It'll use slightly more gas to have to post the timestamp as well, but the abstraction is cleaner.

specs/protocol/block_header.md
```diff
@@ -0,0 +1,12 @@
...
+## Application Header
+
+The application header is a network-agnostic block header. Different [networks](../network/index.md) may wrap the application header in a consensus header, depending on their consensus protocol.
+
+name                  | type       | description
+----------------------|------------|----------------------------------------------------------------------------------------------------------------------------------
+`txCount`             | `uint64`   | Number of [transaction](./tx_format.md#transaction)s in this block.
+`outputMessagesCount` | `uint64`   | Number of [output message](./tx_format.md#outputmessage)s in this block.
+`txRoot`              | `byte[32]` | [Merkle root](./cryptographic_primitives.md#binary-merkle-tree) of [transaction](./tx_format.md#transaction)s in this block.
+`outputMessagesRoot`  | `byte[32]` | [Merkle root](./cryptographic_primitives.md#binary-merkle-tree) of [output message](./tx_format.md#outputmessage)s in this block.
```
Voxelot, line 12: todo: root of output message ID's

@adlerjohn

Voxelot, line 12: todo: add `da_height` for blocks to signal the point of input message inclusion

adlerjohn: Fixed in [430780d](https://github.com/FuelLabs/fuel-specs/pull/405/commits/430780dfa10708c53e77a807681fc6ab7599c3f6)

adlerjohn: Leaving this to #412

Voxelot (approved):

Dentosal (approved):

Issue #286, opened 2022-03-02
Coinbase


adlerjohn:
New tx type: 0-input 1-output, can mint up to the fees collected in the block, must be first tx in block

Have parameter for maturity period (optional, can be left to later)

Implementation Tasks:

- [x] https://github.com/FuelLabs/fuel-specs/issues/421
- [x] https://github.com/FuelLabs/fuel-tx/issues/193
- [x] https://github.com/FuelLabs/fuel-vm/issues/225
- [x] https://github.com/FuelLabs/fuel-core/issues/650

SilentCicero:
Yeah, this one makes things much easier to manage on the fee side.

adlerjohn:
The mint transaction type alone is necessary but not sufficient to specify the coinbase transaction.

Voxelot:
My mistake. To make up for preemptively closing this issue I've opened a PR for specifying coinbase validity rules.

CustomerTeam001:
We'd be happy to look into this for you 
Kindly file a support ticket below 
https://forms.gle/VcaSaDQysAwZBQp77

PR #406, opened 2022-09-15
Add mint transaction type.

adlerjohn:
Partially addresses #286.

Add a new transaction type, `Mint`, that only contains outputs. In the general case, no validation of this transaction is necessary beyond checking for well-formedness. Note that this PR doesn't actually allow mint transactions anywhere, it only defines a new transaction type.

Two applications exist:
1. A coinbase tx, i.e. #286. This will require a block validation rule to check that the first tx is a mint tx and that the values are constrained.
2. Block-level flash loans. https://twitter.com/aeyakovenko/status/1568429252332384257 A mint tx must be paired with an equivalent burn (e.g. a tx with only inputs and no outputs).

### DIFF
--- a/specs/protocol/tx_format.md
+++ b/specs/protocol/tx_format.md
@@ -44,6 +44,7 @@
 enum  TransactionType : uint8 {
     Script = 0,
     Create = 1,
+    Mint = 2,
 }
 ```
 
@@ -52,7 +53,7 @@ enum  TransactionType : uint8 {
 | name   | type                                                                                      | description       |
 |--------|-------------------------------------------------------------------------------------------|-------------------|
 | `type` | [TransactionType](#transactiontype)                                                       | Transaction type. |
-| `data` | One of [TransactionScript](#transactionscript) or [TransactionCreate](#transactioncreate) | Transaction data. |
+| `data` | One of [TransactionScript](#transactionscript), [TransactionCreate](#transactioncreate), or [TransactionMint](#transactionmint) | Transaction data. |
 
 Transaction is invalid if:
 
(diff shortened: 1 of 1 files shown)


specs/protocol/tx_format.md
```diff
@@ -166,6 +167,18 @@ Transaction is invalid if:
 
 Creates a contract with contract ID as computed [here](./identifiers.md#contract-id).
 
+## TransactionMint
```
pixelcircuits, line 170: Add link to this in the TOC at the top

pixelcircuits:
I'm not sure I get how this could be used for flash loans. Are you thinking coins get added to free balance and have to end still in free balance?

Also, I think you could probably go ahead and add the restriction that the `OutputCoins` have to be for the base asset. Having the ability to flash loan any coin (especially if the coin explicitly would not like that ability) opens up potential security concerns for DAOs and on-chain voting

adlerjohn:
> I'm not sure I get how this could be used for flash loans.

See the tweet linked in the OP: https://twitter.com/aeyakovenko/status/1568429252332384257. It could be implemented as the second tx in the block can be a mint tx (the first being the coinbase), for literally any amount of any asset. Then the last tx in the block must burn (i.e. have only inputs and no outputs) the same amount that was minted.

> I think you could probably go ahead and add the restriction that the OutputCoins have to be for the base asset.

Restricting this at the tx format level instead of the validation rules level means it can't be used for the above flash loans.

> Having the ability to flash loan any coin (especially if the coin explicitly would not like that ability) opens up potential security concerns for DAOs and on-chain voting

Good. On-chain voting should always hold assets for at least one block anyways, since flash loans are a thing. This feature also prevents people from abusing single-unit fungible assets to represent non-fungible assets, which would be very wasteful and inefficient.

Note that flash loans aren't the subject of this PR; this PR is only for a general tx format that has several applications.

specs/protocol/tx_format.md
adlerjohn: Fixed in dfadb59

Voxelot (approved):

PR #408, opened 2022-09-26
Improve wording for alu intructions flags

### LINKED ISSUE #316: Unclear behaviour around flags `F_WRAPPING` & `F_UNSAFEMATH`
>If the F_UNSAFEMATH flag is unset, an operation that would have set $err to true is instead a panic.
>If the F_WRAPPING flag is unset, an operation that would have set $of to a non-zero value is instead a panic.

It might be just me, but these make it sound like the user needs to take an action (ie: "unset" the flags) in order to have the panic behaviour. As panic is the default behaviour (and the flags are "unset" to start with), I propose the wording change to something like:

- If the F_UNSAFEMATH flag is set, an operation that would have panicked will instead set `$err` to a non-zero value.
- If the F_WRAPPING flag is set, an operation that would have panicked will instead set `$of` to a non-zero value.

cc @adlerjohn

nfurfaro:
Closes #316

### DIFF
--- a/specs/vm/instruction_set.md
+++ b/specs/vm/instruction_set.md
@@ -118,9 +118,9 @@ then append an additional receipt to the list of receipts, again modifying `tx.r
 
 All these instructions advance the program counter `$pc` by `4` after performing their operation.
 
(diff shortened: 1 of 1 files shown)


otrho (approved):
Yep, now more readable to me.

PR #409, opened 2022-09-27
Add data field to panic receipt

Voxelot:
When the SDK is trying to estimate which contracts should be added to the inputs of a transaction, it is only able to obtain a u8 panic reason from the VM without any additional info about what contract ID was expected to be in the inputs. While the VM knows what contract ID is missing from the inputs, the ABI spec currently doesn't provide any way for the VM to return it.

One way this could be handled is by adding a new 32 byte `data` field to panic receipts, which is what is contained in this PR proposal. A 32 byte data field on panic receipts could also be used for reporting an asset ID when doing ops like `TR` or `TRO`, to indicate which asset has insufficient balance.

This would help unblock https://github.com/FuelLabs/fuels-rs/issues/595

### DIFF
--- a/specs/protocol/abi.md
+++ b/specs/protocol/abi.md
@@ -553,16 +553,18 @@ _Important note:_ For the JSON representation of receipts, we represent 64-bit u
 - `type`: `Panic`.
 - `id`: Hexadecimal string representation of the 256-bit (32-byte) contract ID of the current context if in an internal context. `null` otherwise.
 - `reason`: Decimal string representation of an 8-bit unsigned integer; panic reason.
+- `data`: Hexadecimal string representation of the 256-bit (32-byte) contract ID or asset ID related to the error. `null` otherwise.
 - `pc`: Hexadecimal string representation of a 64-bit unsigned integer; value of register `$pc`.
 - `is`: Hexadecimal string representation of a 64-bit unsigned integer; value of register `$is`.
 
 ```json
 {
-  "type":"Panic",
-  "id":"0x39150017c9e38e5e280432d546fae345d6ce6d8fe4710162c2e3a95a6faff051",
-  "reason":"1",
-  "pc":"0xffffffffffffffff",
-  "is":"0xfffffffffffffffe"
+  "type": "Panic",
+  "id": "0x39150017c9e38e5e280432d546fae345d6ce6d8fe4710162c2e3a95a6faff051",
+  "reason": "1",
+  "data": "0x0000000000000000000000000000000000000000000000000000000000000000",
+  "pc": "0xffffffffffffffff",
+  "is": "0xfffffffffffffffe"
 }
 ```
 
--- a/specs/protocol/abi.md
+++ b/specs/protocol/abi.md
@@ -555,7 +555,7 @@ _Important note:_ For the JSON representation of receipts, we represent 64-bit u
 - `reason`: Decimal string representation of an 8-bit unsigned integer; panic reason.
 - `pc`: Hexadecimal string representation of a 64-bit unsigned integer; value of register `$pc`.
 - `is`: Hexadecimal string representation of a 64-bit unsigned integer; value of register `$is`.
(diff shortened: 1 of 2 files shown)


specs/vm/instruction_set.md
```diff
@@ -99,12 +99,14 @@ Some instructions may _panic_, i.e. enter an unrecoverable state. Additionally,
...
-| `pc`   | `uint64`      | Value of register `$pc`.                                                  |
-| `is`   | `uint64`      | Value of register `$is`.                                                  |
+| name     | type          | description                                                                    |
+|----------|---------------|--------------------------------------------------------------------------------|
+| `type`   | `ReceiptType` | `ReceiptType.Panic`                                                            |
+| `id`     | `byte[32]`    | Contract ID of current context if in an internal context, zero otherwise.      |
+| `pc`     | `uint64`      | Value of register `$pc`.                                                       |
+| `is`     | `uint64`      | Value of register `$is`.                                                       |
+| `reason` | `uint8`       | The panic reason identifier                                                    |
+| `data`   | `byte[32]`    | ContractId or AssetID associated with the panic if applicable, zero otherwise. |
```
adlerjohn, line 109: Hmmm. So two things:

1. Including a canonical panic reason would make the specs way more verbose. It was actually removed in #256!
2. `data` isn't really needed in the _canonical_ receipt. Especially given that the primary usecase of the requested feature is wallets simulating txs, where the simulated txs don't actually end up on-chain.

I would simply omit any changes to the canonical receipt and only keep the JSON receipt changes.

adlerjohn (changes requested):

specs/vm/instruction_set.md
Voxelot: So in essence we can add metadata fields to the panic receipt as an implementation detail, as long as it's not included into the blockchain state or vm/ivg logic?

adlerjohn: `:yes:`

adlerjohn: It might be good to distinguish between optional and non-optional fields.

specs/protocol/abi.md
```diff
@@ -553,16 +553,18 @@ _Important note:_ For the JSON representation of receipts, we represent 64-bit u
 - `type`: `Panic`.
 - `id`: Hexadecimal string representation of the 256-bit (32-byte) contract ID of the current context if in an internal context. `null` otherwise.
 - `reason`: Decimal string representation of an 8-bit unsigned integer; panic reason.
+- `data`: Hexadecimal string representation of the 256-bit (32-byte) contract ID or asset ID related to the error. `null` otherwise.
```
adlerjohn, line 556: Instead of `null`, could this field be made optional?

vlopes11: My concern about `null` is it is nouveau to the specs. The serialization of this new variant should probably be a boolean byte with a flag on whether or not the field is there, and the following bytes optionally with the contents.

I would also avoid creating fields that might contain different meanings. In this case, we should probably have two optional fields - one for the contract id, the other for the asset id. Having the information mixed might lead to complications downstream.

adlerjohn: `null` is used elsewhere though, see `id` above.

vlopes11: True; interestingly enough, we still serialize the (zeroed) fields in the canonical serialization:

https://github.com/FuelLabs/fuel-tx/blob/0cb216da4fa2a823caa402f1973594cda4406f9b/src/receipt/receipt_std.rs#L75-L83

If we just extend the same strategy then we should be fine. I don't think this discrepancy between abi and memory serialization will ever be an issue

specs/vm/instruction_set.md
Voxelot: Removed from VM rules

specs/protocol/abi.md
Voxelot: Updated this to `contractId`, we can add `assetId` later if actually needed.

adlerjohn (approved):

PR #411, opened 2022-10-04
OutputMessage format change to messageID

Voxelot:
The current fields on message outputs weren't useful for enabling withdrawal proofs. Since the messageID is inclusive of all other message fields (derivable from the receipt), this PR simplifies the spec to make messageID the only field on an output message.

Since the block header has a merkle root of messageIDs, storing the message id on the output simplifies the process of generating roots and proofs for messages.

### DIFF
--- a/specs/protocol/tx_format.md
+++ b/specs/protocol/tx_format.md
@@ -349,14 +349,13 @@ The state root `stateRoot` is the root of the [SMT](./cryptographic_primitives.m
 
 ### OutputMessage
 
-| name        | type       | description                                   |
-|-------------|------------|-----------------------------------------------|
-| `recipient` | `byte[32]` | The address of the message recipient.         |
-| `amount`    | `uint64`   | Amount of base asset coins sent with message. |
+| name        | type       | description                                                                              |
+|-------------|------------|------------------------------------------------------------------------------------------|
+| `messageID` | `byte[32]` | The messageID as described [here](../protocol/identifiers.md#message-id). 0x00 if unset. |
 
-Note: when signing a transaction `recipient` and `amount` are set to zero.
+Note: when signing a transaction `messageID` is set to zero.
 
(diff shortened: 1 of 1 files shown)


adlerjohn (commented):
Leaving general notes of the message root in the header requires messages to 1) be uniquely identifiable and 2) include a commitment to their payload/data, for the bridge contract To be done in #412. For this PR, the canonical receipt and output message can include whatever data is easier for us (the client). One condition is at least that receipts need to include enough information to order the output messages totally, i.e. _at least_ the output index used by the `SMO` executed (directly or indirectly).

Voxelot:
After understanding the problem better I'm leaning towards just closing the PR. But in summary the main difference comes down to this:

- We order messages based on receipts at the cost of having to load all the receipts for generating proofs. With this approach, output message fields are basically useless to the client and we can just leave the specs as is since the receipts have all the data we need.
- We order messages based on a total ordering of tx idx + output idx. In order to avoid loading receipts, the message outputs should include all the fields necessary for computing a messageId that commits to fields like the data (digest) and recipient. Ultimately this would just save some I/O on the client so we don't have to touch receipts but create some hypothetical risks for fraud proofs which may not even have a message root. It would also involve changing the way message IDs are verified in the contracts to use data digest instead of data. Since this really only affects read-only APIs to serve up withdrawal proofs, it's probably not worth the effort to change the tx format at this stage.

adlerjohn:
If we're going for a "avoid looking too far into the future" approach...then we could just remove output messages entirely?!

Voxelot:
> If we're going for a "avoid looking too far into the future" approach...then we could just remove output messages entirely?!

Doesn't make much of a difference to the client whether these outputs exist assuming receipts are sufficient.

xgreenx:
The `MessageOutput` is removed entirely in the https://github.com/FuelLabs/fuel-specs/pull/453

Issue #395, opened 2022-08-03
Storage tree differentiate between default and non-default


adlerjohn:
In light of the most recent exploit on Nomad (<https://twitter.com/samczsun/status/1554260106107179010>), there's a need to differentiate between default and non-default leaves when querying the contract storage tree.

## Current Design

The default value of a leaf in the SMT is empty (i.e. zero bytes). However, when reading from storage, `0` is returned if the leaf is empty. This is lossy since it's discarding information.

https://github.com/FuelLabs/fuel-specs/blob/d19cd0ef6fbac7619d5d04b1f00e869ea78593fc/specs/vm/instruction_set.md?plain=1#L1444

Moreover, there's no way to reset a storage slot; writing `0` is not the same thing as writing an empty value.

## Proposed Changes

1. Add a new instruction, to clear a storage slot. The instruction resets the given storage slot to the default value (i.e. zero bytes).
2. Add an additional register parameter to `SRW` and `SRWQ`, that will be set to a Boolean for whether the given storage slot was empty or not.

sezna:
> Add an additional register parameter to SRW and SRWQ, that will hold a flag for whether the given storage slot was empty or not.

Does "empty" here mean zeroed out? Because that is also lossy. Or will the VM keep track of initialized (occupied) vs uninitialized (cleared) slots in a separate bitfield?

adlerjohn:
The SMT already differentiates between "empty" and "32 bytes of zeroes," and therefore so does the VM implicitly. The proposed changes simply expose this in the instruction set.

SwayStar123:
this is quite poggers if i say so myself

SwayStar123:
> Add an additional register parameter to SRW and SRWQ, that will hold a flag for whether the given storage slot was empty or not.
Not familiar with what SRW and SRWQ are, I thought of a flag type system before to solve this problem (like the sign on a signed int) but then I realized that would tack on a 1 bit cost to every single slot used, might add up quickly? Im assuming thats how SRW works aswell, do you not see this as a potential problem?

adlerjohn:
There's actually no need for any changes to the SMT because the SMT actually _already differentiates between a default leaf and a zero-value non-default leaf_. The extra register parameter to `SRW` and `SRWQ` would simply be set to zero or one depending on whether the slot was default or non-default.

Voxelot:
@adlerjohn should we close this issue now? The spec changes are in and the impl is done (pending sequential storage slot access).

[...]

PR #415, opened 2022-10-12
Add clear sequential word command

mitch-fuel:
Part of Issue #395.

Add clear sequential word command.


specs/vm/instruction_set.md
```diff
@@ -1417,6 +1418,22 @@ This modifies the `balanceRoot` field of the appropriate output.
...
+| Syntax      | `scwq $rA, $rB`                                                               |
+| Encoding    | `0x00 rA rB - -`                                                              |
+| Notes       |                                                                               |
+
+Panic if:
+
+- `$rA + 32 * rB` overflows
+- `$rA + 32 * rB > VM_MAX_RAM`
+- `$fp == 0` (in the script context)
+
```
Voxelot, line 1436: Should we also set `rB` here to indicate whether the initial key actually existed?

specs/vm/instruction_set.md
```diff
@@ -1417,6 +1418,22 @@ This modifies the `balanceRoot` field of the appropriate output.
...
+| Description | A sequential series of 32 bytes is cleared from the current contract's state. |
+| Operation   | ```STATE[MEM[$rA, 32 * rB]] = None;```                                        |
+| Syntax      | `scwq $rA, $rB`                                                               |
+| Encoding    | `0x00 rA rB - -`                                                              |
+| Notes       |                                                                               |
+
+Panic if:
+
+- `$rA + 32 * rB` overflows
+- `$rA + 32 * rB > VM_MAX_RAM`
```
Voxelot, line 1434: Since we are clearing data from the database (state) instead of VM memory, it seems like this check isn't applicable.

specs/vm/instruction_set.md
```diff
@@ -1417,6 +1418,22 @@ This modifies the `balanceRoot` field of the appropriate output.
 `messageID` is added to the `OutputMessage` Merkle tree as part of block header.
 TODO: document output messages merkle tree construction and maintenance and link here
 
+### SCWQ: State clear sequential 32 byte slots
+
+|             |                                                                               |
+|-------------|-------------------------------------------------------------------------------|
+| Description | A sequential series of 32 bytes is cleared from the current contract's state. |
+| Operation   | ```STATE[MEM[$rA, 32 * rB]] = None;```                                        |
```
Voxelot, line 1426: I'm assuming the intention is to provide a starting key and delete the next `n` entries, rather than providing an array of keys to bulk delete from the db?

```suggestion
| Operation   | ```STATE[MEM[$rA, 32], 32  * $rB] = None;```                                        |
```

specs/vm/instruction_set.md
mitch-fuel: Those checks exist for SWW and SWWQ. Should those also not exist? Or are you saying that they are getting their input values from VM Memory so the check is necessary there but our input value is already known?

specs/vm/instruction_set.md
```diff
@@ -1417,6 +1418,22 @@ This modifies the `balanceRoot` field of the appropriate output.
...
+| Syntax      | `scwq $rA, $rB`                                                               |
+| Encoding    | `0x00 rA rB - -`                                                              |
+| Notes       |                                                                               |
+
+Panic if:
+
+- `$rA + 32 * rB` overflows
```
Voxelot, line 1433: [nit] inconsistent formatting:`$rA` has `$` but `rB` does not

specs/vm/instruction_set.md
Voxelot: Yeah the difference is that SWW and SWWQ are copying bytes from db into memory, and we need to ensure the memory range withing the VM is valid. 

Clearing contiguous state is different, because we are specifying data ranges inside of the DB instead of VM memory. I edited a previous comment to help clarify the intent of how I think this would be used but it might've been resolved before you saw it:

> The reason for this being that when data in sway is stored to the db, it is packed into contiguous 32 byte storage slots with an ascendingly increasing storage key. So if you wanted to remove an entire struct from the db in one operation, you could just provide the starting address and the number of slots that were used to store it.

https://github.com/FuelLabs/fuel-specs/pull/415#discussion_r993932158

Voxelot: Also realized, you meant SRW and SRWQ right? For SWW and SWWQ, it doesn't seems like they have any `32 * d` type ops yet.

specs/vm/instruction_set.md
```diff
@@ -1417,6 +1418,22 @@ This modifies the `balanceRoot` field of the appropriate output.
...
+| Description | A sequential series of 32 bytes is cleared from the current contract's state. |
+| Operation   | ```STATE[MEM[$rA, 32], 32  * $rB] = None;```                                        |
+| Syntax      | `scwq $rA, $rB`                                                               |
+| Encoding    | `0x00 rA rB - -`                                                              |
+| Notes       |                                                                               |
+
+Panic if:
+
+- `$rA + 32 * $rB` overflows
+- `$rA + 32 * $rB > VM_MAX_RAM`
```
Voxelot, line 1434: You resolved my previous comment about this, but I don't see any replies or updates to the PR which resolve this?

specs/vm/instruction_set.md
adlerjohn: Unlike x86, we don't re-use registers anywhere in the instruction set. So if we want to return something, we should use a third register.

specs/vm/instruction_set.md
```diff
@@ -1417,6 +1418,22 @@ This modifies the `balanceRoot` field of the appropriate output.
...
+|-------------|-------------------------------------------------------------------------------|
+| Description | A sequential series of 32 bytes is cleared from the current contract's state. |
+| Operation   | ```STATE[MEM[$rA, 32], 32  * $rB] = None;```                                        |
+| Syntax      | `scwq $rA, $rB`                                                               |
+| Encoding    | `0x00 rA rB - -`                                                              |
+| Notes       |                                                                               |
+
+Panic if:
+
+- `$rA + 32 * $rB` overflows
```
adlerjohn, line 1433: Wait, you don't need to multiple by `$rB` here. `MEM[$rA, 32]` holds the 32 byte slot index and that's it.

specs/vm/instruction_set.md
Voxelot: yeah I was assuming we'd shift the current use of rB into rC to be more consistent with the other recent changes

specs/vm/instruction_set.md
Voxelot: I made the same comment here: https://github.com/FuelLabs/fuel-specs/pull/415#discussion_r993929556

mitch-fuel: fixed

specs/vm/instruction_set.md
mitch-fuel: fixed

Voxelot (approved):
lgtm

Although should we return something else via $rB if you request more slots to be deleted than exist?

Another implementation question I have is whether we have any criteria for what is considered a sequential slot. Do we intend to be able to put in a starting address of 0x00 and set rC to u64::MAX to nuke all the state of the contract? Assuming this is mainly intended to service the way that sway packs structs into db slots, can we assume that all related slot keys only have a difference of one between them? 

For example let's say we have the following setup:

struct A => [slot_1, slot_2, slot_3]
struct B => [slot_10, slot_11_slot_13]
struct C => [slot_200, slot_201, slot_202]

If I call 
`scwq(slot_1, $rB, 100)`

should it start deleting at slot_1 and keep going for the next 100 entries it finds in the linked list of slot keys? Or should it stop when there's a "gap" between slot key groupings which may indicate different structs? Or should stop deleting once slot_100 is reached even though none of the slots in between were used?

specs/vm/instruction_set.md
```diff
@@ -1417,6 +1418,25 @@ This modifies the `balanceRoot` field of the appropriate output.
...
+| Notes       |                                                                               |
+
+Panic if:
+
+- `$rA + 32` overflows
+- `$rA + 32 > VM_MAX_RAM`
+- `$rB` is a [reserved register](./main.md#semantics)
+- `$fp == 0` (in the script context)
+
+Register `rB` will be populated with `false` if the first word was already unset (default) and `true` if the value is set.
```
adlerjohn, line 1438:
```suggestion
Register `rB` will be set to `false` if the first storage slot was already unset (default) and `true` if the slot was set.
```

adlerjohn:
> Do we intend to be able to put in a starting address of 0x00 and set rC to u64::MAX to nuke all the state of the contract?

Sure, why not. It would be expensive and cause a running out of gas immediately, but sure.

> Assuming this is mainly intended to service the way that sway packs structs into db slots, can we assume that all related slot keys only have a difference of one between them?

Yes.

I'm not 100% what you mean by linked list of keys, but if you want to clear 100 storage slots starting at storage slot 0, you'd clear storage slots `[0, 100)` in key in the SMT. The `STATE[MEM[$rA, 32], 32 * $rC]` line should be unambiguous in what's meant.

mitch-fuel:
After the last bit of conversation are we looking for any more clarifications? I think it looks like concerns are resolved now.

Voxelot:
> I'm not 100% what you mean by linked list of keys, but if you want to clear 100 storage slots starting at storage slot 0, you'd clear storage slots [0, 100) in key in the SMT. The STATE[MEM[$rA, 32], 32 * $rC] line should be unambiguous in what's meant.

Sounds good to me, I'm satisfied with the level of clarity here then. It's just a little counter-intuitive from a db perspective, since usually we get the next n entries after the starting key, rather than defining absolute ranges since we don't know how many records that would be. 

This could come up in gas profiling. For instance, clearing slots `[0, 100)` may do up to 100 deletions, or it could perform zero, and we don't really know in advance how much I/O will be performed and will have to charge the worst case every time.

> After the last bit of conversation are we looking for any more clarifications? I think it looks like concerns are resolved now.

Yeah I think it's clear enough from my perspective.

Voxelot (approved):

PR #416, opened 2022-10-13
Update the ABI spec with the new JSON ABI format

mohammadfawaz:


sezna (approved):
approving but @digorithm should take a look before this goes in

digorithm:
Taking another look now 👀

digorithm (approved):
Very nice; thanks for doing this, @mohammadfawaz!

PR #422, opened 2022-10-20
Write sequential storage slots with previous status

mitch-fuel:
Part of issue #395.

Will allow for writing of sequential storage slots and will populate rB with whether the first slot was previously set.


adlerjohn (approved):
Seems reasonable.

PR #424, opened 2022-10-26
set rB to false if any storage slot in the requested range is unset

Voxelot:


adlerjohn (approved):

adlerjohn:
For posterity, this change was done in order to avoid the case of upgradable contracts not having guarantees that storage is set/unset consistently by the compiler.

PR #425, opened 2022-10-30
Move specs into an `mdbook`. Add CI for deploying.

### LINKED ISSUE #387: Consider hosting the spec as an mdbook?
Perhaps it's worth considering hosting the spec under an mdbook?

Some benefits that come to mind:

- Easier to navigate for folk unfamiliar with GitHub.
- The chapter column provides a nice overview with subheaders etc.
- Left and right keys for navigating through pages.
- Supports simpler, real-time search.
- Has a community providing external tooling for conversion to PDF, ePub, etc, handy for offline or mobile reading.

For the most part the file layout would almost be the same and folk would still be welcome to peruse by clicking through GitHub links if they prefer.

I remember there being some brief discussion around keeping this as raw GitHub markdown but I don't recall the reasoning :thinking:

mitchmindtree:
I was reading through the spec today and figured I'd address this at the same time!

Closes #387.

Screenshot from 2022-10-30 17-48-28

I broke up `tx_format.md` a little into subchapters to make it a little easier to navigate from the sidebar. We could probably do the same for the VM Instruct Set categories in a follow-up PR.


adlerjohn:
image

adlerjohn (dismissed):

mohammadfawaz (approved):
Awesome 🥳

PR #427, opened 2022-10-30
Annotate VM instructions with their effects

### LINKED ISSUE #426: Document effects specification for VM instructions
Would we want to add “effects specification” to the VM instructions [spec](https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/instruction_set.md)? I’m thinking about an ‘effects’ row in those small summary tables. The values would be
* `no external effect` or `pure` or something similar;
* `storage write`;
* `storage read`;
* `external call`,
* ...

Motivation: the PRs https://github.com/FuelLabs/sway/pull/3168 and https://github.com/FuelLabs/sway/pull/3648 introduce the CEI pattern analysis of effects after external contract calls. To do that, I needed to understand what instructions produce effects and it would be great to have those documented in the specs.

anton-trunov:
I annotated the instructions with their effects to resolve #426.
Rendered file: https://github.com/FuelLabs/fuel-specs/blob/anton-trunov/vm-instructions-effects-specification/specs/vm/instruction_set.md


specs/vm/instruction_set.md
```diff
@@ -937,6 +939,7 @@ Panic if:
 | Operation   | ```$rA = MEM[$rB, $rD] == MEM[$rC, $rD];``` |
 | Syntax      | `meq $rA, $rB, $rC, $rD`                    |
 | Encoding    | `0x00 rA rB rC rD`                          |
+| Effects     | Storage read                                |
```
mohammadfawaz, line 942: Ah so instructions that access `MEM[..]` do not actual read/write storage. They only access VM memory so they technically have no permanent side effects.

I believe the only instructions that access storage are `srw`, `srwq`, `sww`, `swwq`, `bal`, and `scwq` (Just added a we don't support yet in the compiler)

anton-trunov: This is really useful! Thank you

specs/vm/instruction_set.md
```diff
@@ -131,6 +131,7 @@ If the [`F_WRAPPING`](./main.md#flags) flag is set, an operation that would have
 | Operation   | ```$rA = $rB + $rC;``` |
 | Syntax      | `add $rA, $rB, $rC`    |
 | Encoding    | `0x00 rA rB rC -`      |
+| Effects     | No effects             |
```
mitchmindtree, line 134: Perhaps we can assume "No effects" by default, and only add an `Effects` row for "impure" instructions?

mitchmindtree (commented):
Nice, I really like this idea!

Perhaps It could also be useful to have a section under `vm/index.md` that lists the set of possible effects Fuel VM instructions can have? I can see it being useful for VM implementors to see all of these in one place, knowing that otherwise all instructions are "pure".

specs/vm/instruction_set.md
anton-trunov: On one hand, I don't have a preference here, on the other hand, if the effects row is always present it makes the instruction summaries more uniform and makes sure we didn't forget to annotate an instruction or two :)

As you suggested in the other comment above, we should definitely enumerate the set of all possible effects in `vm/index.md` to make it uniform.

anton-trunov: Addressed review suggestion in a new commit, I think it's ready for review now

mitchmindtree (approved):
Nice! I haven't gone through all the instructions to check if any have been missed, but what has been introduced in this PR looks good :)

anton-trunov:
@mitchmindtree Thanks!

@mohammadfawaz Are you ok with the changes?

mohammadfawaz (approved):

mohammadfawaz:
> @mitchmindtree Thanks!
> 
> @mohammadfawaz Are you ok with the changes?

Yup looks good! Thanks.

PR #433, opened 2022-11-15
Split `abi.md` into multiple files

AlicanC:
This PR splits the "ABI" section into three parts for easier access:
image

The actual content is left untouched.


mohammadfawaz:
Will you make sure that all links to the old `abi.md` file from the specs, the Sway Book, and the fuel book (if any) are updated?

AlicanC:
> Will you make sure that all links to the old `abi.md` file from the specs, the Sway Book, and the fuel book (if any) are updated?

How about a redirect? https://github.com/FuelLabs/fuel-specs/pull/433/commits/70a6936523b60142869cc1764c902fb37057da2a

AlicanC:
Leaving a redirect for every change like this sounds like it would introduce a lot of clutter in the long term, but this also makes users' bookmarks not break, etc.

mohammadfawaz (approved):

PR #437, opened 2022-12-02
Bring back argument encoding

mohammadfawaz:
This section was omitted when the specs were rewritten as a "Book"

### DIFF
--- a/src/protocol/abi/argument_encoding.md
+++ b/src/protocol/abi/argument_encoding.md
@@ -1,0 +1,245 @@
+# Argument Encoding
+
+When crafting transaction script data, you must encode the arguments you wish to pass to the script.
+
+## Types
+
+These are the available types that can be encoded in the ABI:
+
+- Unsigned integers:
+  - `u8`, 8 bits.
+  - `u16`, 16 bits.
+  - `u32`, 32 bits.
+  - `u64`, 64 bits.
+- Boolean: `bool`, either `0` or `1` encoded identically to `u8`.
+- B256: `b256`, arbitrary 256-bits value.
+- Address : `address`, a 256-bit (32-byte) address.
+- Fixed size string
+- Array
+- Sum types
+- Structs
+
+These types are encoded in-place. Here's how to encode them. We define `enc(X)` the encoding of the type `X`.
+
+## Unsigned Integers
+
+`u<M>` where `M` is either 8, 16, 32, or 64 bits.
+
+`enc(X)` is the big-endian representation of `X` left-padded with zero-bytes. Total length must be 8 bytes.
+
+> **Note:** since all integer values are unsigned, there is no need to preserve the sign when extending/padding; padding with only zeroes is sufficient._
+
+**Example:**
+
+Encoding `42` yields: `0x000000000000002a`, which is the binary representation of the decimal number `42`, right-aligned to 8 bytes.
+
+## Boolean
+
+`enc(X)` is `0` if `X` is false or `1` if `X` is true, left-padded with zero-bytes. Total length must be 8 bytes. Similar to the `u8` encoding.
+
+**Example:**
+
+Encoding `true` yields:
+
+```text
+0x0000000000000001
+```
+
+## B256
+
+`b256` is a fixed size bit array of length 256. Used for 256-bit hash digests and other 256-bit types. It is encoded as-is.
+
+**Example:**
+
+Encoding `0xc7fd1d987ada439fc085cfa3c49416cf2b504ac50151e3c2335d60595cb90745` yields:
+
+```text
+0xc7fd1d987ada439fc085cfa3c49416cf2b504ac50151e3c2335d60595cb90745
+```
+
+## Address
+
+A 256-bit (32-byte) address, encoded in the same way as a `b256` argument: encoded as-is.
+
+**Example:**
+
+Encoding `0xc7fd1d987ada439fc085cfa3c49416cf2b504ac50151e3c2335d60595cb90745` yields:
+
+```text
+0xc7fd1d987ada439fc085cfa3c49416cf2b504ac50151e3c2335d60595cb90745
+```
+
+## Array
+
+An array of a certain type `T`, `[T; n]`, where `n` is the length of the array.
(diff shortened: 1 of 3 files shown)


src/protocol/abi/argument_encoding.md
```diff
@@ -0,0 +1,245 @@
...
+  - `u8`, 8 bits.
+  - `u16`, 16 bits.
+  - `u32`, 32 bits.
+  - `u64`, 64 bits.
+- Boolean: `bool`, either `0` or `1` encoded identically to `u8`.
+- B256: `b256`, arbitrary 256-bits value.
+- Address : `address`, a 256-bit (32-byte) address.
+- Fixed size string
+- Array
+- Sum types
```
otrho, line 19: Do we call them sum types elsewhere?  It's a bit of a language nerd term, where devs will just use Enum.

src/protocol/abi/argument_encoding.md
```diff
@@ -0,0 +1,245 @@
...
+
+`u<M>` where `M` is either 8, 16, 32, or 64 bits.
+
+`enc(X)` is the big-endian representation of `X` left-padded with zero-bytes. Total length must be 8 bytes.
+
+> **Note:** since all integer values are unsigned, there is no need to preserve the sign when extending/padding; padding with only zeroes is sufficient._
+
+**Example:**
+
+Encoding `42` yields: `0x000000000000002a`, which is the binary representation of the decimal number `42`, right-aligned to 8 bytes.
```
otrho, line 34: OTOH, is 'big-endian' too nerdy and 'right aligned' better?  Right aligned sounds weird to me, but I'm a language nerd.

src/protocol/abi/argument_encoding.md
```diff
@@ -0,0 +1,245 @@
...
+
+Calling `bar` with `MySumType::Y(42)` yields:
+
+```plaintext
+0x
+0000000000000001 // The discriminant of the chosen enum, in this case `1`.
+0000000000000000 // Left padding
+0000000000000000 // Left padding
+0000000000000000 // Left padding
+000000000000002a // `42` encoded as u64
```
otrho, line 217: Does this correlate at all to how we encode them in Sway itself?  I'm pretty sure that the padding goes after the value in the memory layout in the compiler... but is that irrelevant to this ABI encoding stuff?

Also does the empty value for all-unit variants apply here or not?  I thought not, but... maybe?

src/protocol/abi/argument_encoding.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Array
+
+An array of a certain type `T`, `[T; n]`, where `n` is the length of the array.
+
+Arrays in Sway have a fixed-length which is known at compile time. This means the ABI encoding for arrays also happens in-place, with no need to account for dynamic sizing.
+
+The encoding for the array contains, in order, the encoding of each element in `[T; n]`, recursively following the encoding for the type `T`.
+
+For instance, consider the function signature `my_func(bool,[u64; 2])` with the values `(true,[1, 2])`.
```
mitchmindtree, line 80:
```suggestion
For instance, consider the function signature `my_func(bool, [u64; 2])` with the values `(true, [1, 2])`.
```

src/protocol/abi/argument_encoding.md
mitchmindtree, line 34:
```suggestion
Encoding `42` yields: `0x000000000000002a`, which is the hexadecimal representation of the decimal number `42`, right-aligned to 8 bytes.
```

src/protocol/abi/argument_encoding.md
```diff
@@ -0,0 +1,245 @@
...
+
+abi MyContract {
+    fn foo(a: u64);
+    fn bar(a: InputStruct);
+} {
+    fn baz(a: ()) { }
+}
+```
+
+Calling `bar` with `InputStruct{field_1: true, field_2: 5}` yields:
```
mitchmindtree, line 131:
```suggestion
Calling `bar` with `InputStruct { field_1: true, field_2: 5 }` yields:
```

src/protocol/abi/argument_encoding.md
```diff
@@ -0,0 +1,245 @@
...
+
+abi MyContract {
+    fn foo(a: u64);
+    fn bar(a: InputStruct);
+} {
+    fn baz(a: ()) { }
+}
+```
+
+Calling `bar` with `InputStruct{field_1: true, field_2: [1,2]}` yields:
```
mitchmindtree, line 156:
```suggestion
Calling `bar` with `InputStruct{ field_1: true, field_2: [1,2] }` yields:
```

src/protocol/abi/argument_encoding.md
adlerjohn: I would prefer big-endian myself.

src/protocol/abi/argument_encoding.md
adlerjohn: porque no los dos?

adlerjohn (approved):
Ok assuming this is a 1:1 copy of the old text. But can be improved as per comments.

mitchmindtree (approved):
> This section was omitted when the specs were rewritten as a "Book"

Ahh good catch! Dunno how I missed such a large section :monocle_face: 

A few small suggestions, but I'll approve as it's probably more important to get this section back in the spec before improving parts of it. Happy to re-approve if you want to address any of these though :+1:

mohammadfawaz:
Will create a follow up PR addressing all the comments.

src/protocol/abi/argument_encoding.md
mohammadfawaz: We have [this code](https://github.com/FuelLabs/sway/blob/4d5fea688f47617f9644deae387571e3140f5f50/sway-core/src/asm_generation/from_ir.rs#L254) in `from_ir.rs`. I remember aligning that with the spec. Same for all-unit variants and there are tests to make sure interfacing with the SDK is correct. We've made sure in the past that the memory layout is exactly the same internally and in the ABI.

PR #442, opened 2022-12-09
specify coinbase validity rules

Voxelot:
Adds validity rules for coinbase transactions

fixes: https://github.com/FuelLabs/fuel-specs/issues/286

### DIFF
--- a/src/protocol/tx_validity.md
+++ b/src/protocol/tx_validity.md
@@ -190,7 +190,7 @@ The coinbase is a mechanism for block creators to convert fees into spendable UT
 
 In order for a coinbase to be valid:
 
-1. It must issued in the form of a [Mint](../protocol/tx_format/transaction.md#TransactionMint) transaction.
+1. It must be a [Mint](../protocol/tx_format/transaction.md#TransactionMint) transaction.
 2. The coinbase must be the first transaction within a block.
 3. The total output value of the coinbase cannot exceed the total amount of fees processed from all other transactions within the same block.
 4. The `asset_id` for coinbase outputs must match the `asset_id` that fees are paid in (`asset_id == 0`).
--- a/src/protocol/tx_validity.md
+++ b/src/protocol/tx_validity.md
@@ -184,13 +184,13 @@ If change outputs are present, they must have:
 
 Transaction processing is completed by removing spent UTXOs from the state and adding created UTXOs to the state.
 
-### Coinbase
+### Coinbase Transaction
 
-The coinbase is a mechanism for block creators to convert fees into spendable UTXOs.
+The coinbase transaction is a mechanism for block creators to convert fees into spendable UTXOs.
 
-In order for a coinbase to be valid:
+In order for a coinbase transaction to be valid:
 
 1. It must be a [Mint](../protocol/tx_format/transaction.md#TransactionMint) transaction.
-2. The coinbase must be the first transaction within a block, even if there are no other transactions in the block and the fee is zero.
-3. The total output value of the coinbase cannot exceed the total amount of fees processed from all other transactions within the same block.
-4. The `asset_id` for coinbase outputs must match the `asset_id` that fees are paid in (`asset_id == 0`).
+2. The coinbase transaction must be the first transaction within a block, even if there are no other transactions in the block and the fee is zero.
+3. The total output value of the coinbase transaction cannot exceed the total amount of fees processed from all other transactions within the same block.
+4. The `asset_id` for coinbase transaction outputs must match the `asset_id` that fees are paid in (`asset_id == 0`).
--- a/src/protocol/tx_validity.md
+++ b/src/protocol/tx_validity.md
@@ -183,3 +183,14 @@ If change outputs are present, they must have:
 ### State Changes
 
 Transaction processing is completed by removing spent UTXOs from the state and adding created UTXOs to the state.
+
+### Coinbase
+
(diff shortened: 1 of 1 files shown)


src/protocol/tx_validity.md
```diff
@@ -183,3 +183,14 @@ If change outputs are present, they must have:
...
+### Coinbase
+
+The coinbase is a mechanism for block creators to convert fees into spendable UTXOs.
+
+In order for a coinbase to be valid:
+
+1. It must issued in the form of a [Mint](../protocol/tx_format/transaction.md#TransactionMint) transaction.
+2. The coinbase must be the first transaction within a block.
+3. The total output value of the coinbase cannot exceed the total amount of fees processed from all other transactions within the same block.
+4. The `asset_id` for coinbase outputs must match the `asset_id` that fees are paid in (`asset_id == 0`).
```
xgreenx, line 196:
```suggestion
4. The `asset_id` for coinbase outputs must match the `asset_id` that fees are paid in (`asset_id == 0` by default).
```

src/protocol/tx_validity.md
```diff
@@ -183,3 +183,14 @@ If change outputs are present, they must have:
...
 Transaction processing is completed by removing spent UTXOs from the state and adding created UTXOs to the state.
+
+### Coinbase
+
+The coinbase is a mechanism for block creators to convert fees into spendable UTXOs.
+
+In order for a coinbase to be valid:
+
+1. It must issued in the form of a [Mint](../protocol/tx_format/transaction.md#TransactionMint) transaction.
+2. The coinbase must be the first transaction within a block.
```
xgreenx, line 194:
```suggestion
2. The coinbase must be the first transaction within a block, even if there are no other transactions in the block and the fee is zero.
```

xgreenx (approved):

src/protocol/tx_validity.md
adlerjohn: I wouldn't say this is a default; rather, it's a requirement that the asset ID of the coinbase tx output MUST be something. That something could change later, but it's still a requirement.

src/protocol/tx_validity.md
```diff
@@ -183,3 +183,14 @@ If change outputs are present, they must have:
...
 
 Transaction processing is completed by removing spent UTXOs from the state and adding created UTXOs to the state.
+
+### Coinbase
+
+The coinbase is a mechanism for block creators to convert fees into spendable UTXOs.
+
+In order for a coinbase to be valid:
+
+1. It must issued in the form of a [Mint](../protocol/tx_format/transaction.md#TransactionMint) transaction.
```
adlerjohn, line 193:
```suggestion
1. It must be a [Mint](../protocol/tx_format/transaction.md#TransactionMint) transaction.
```

adlerjohn (commented):
Would it be good to use "coinbase transaction" here? I imagine that some people may describe the output of the coinbase transaction as the coinbase output. Making the explicit distinction between the two can avoid confusion.

src/protocol/tx_validity.md
xgreenx: It is how it works now. We can make it optional, but it will require patching the existing logic, and we need to keep it in mind during work with block. 

If the most often case for us is non-zero coinbase transactions, then making it optional may not be worth it because the more complex validation rules could cause bugs in the code.

src/protocol/tx_validity.md
xgreenx: I meant that the default asset to pay fee is `0`=) But it can be different for different networks in the future(I don't know why, but who knows)

src/protocol/tx_validity.md
adlerjohn: Alright. Both approaches are viable, so whatever works better for you guys.

adlerjohn:
There's a few outstanding comments @Voxelot, then should be ready for re-review.

src/protocol/tx_validity.md
Voxelot: closing this thread as we use asset_id = 0 widely throughout the specs to refer to the base asset. If we want to make this more configurable that would be a larger, separate change.

Voxelot:
> Would it be good to use "coinbase transaction" here? I imagine that some people may describe the output of the coinbase transaction as the coinbase output. Making the explicit distinction between the two can avoid confusion.

It seems that your comment was detached from the diff, so I just updated any references of coinbase to coinbase transaction.

Lmk if there's any other outstanding comments that need consideration. cc @adlerjohn

src/protocol/tx_validity.md
```diff
@@ -183,3 +183,14 @@ If change outputs are present, they must have:
...
+### Coinbase Transaction
+
+The coinbase transaction is a mechanism for block creators to convert fees into spendable UTXOs.
+
+In order for a coinbase transaction to be valid:
+
+1. It must be a [Mint](../protocol/tx_format/transaction.md#TransactionMint) transaction.
+2. The coinbase transaction must be the first transaction within a block, even if there are no other transactions in the block and the fee is zero.
+3. The total output value of the coinbase transaction cannot exceed the total amount of fees processed from all other transactions within the same block.
+4. The `asset_id` for coinbase transaction outputs must match the `asset_id` that fees are paid in (`asset_id == 0`).
```
adlerjohn, line 196: Thoughts: we should have a parameters page, and make the base asset ID a parameter.

adlerjohn (approved):

Issue #111, opened 2021-02-23
Compressed Address Registries


SilentCicero:
## Motivation
For security, we are enforcing a 32 byte address system. While this increases security, this sized address and even the Ethereum 20 byte address format is quite large for normal day-to-day usage. With the cost of bytes increasing in price on Ethereum, every bit of savings can increase overall transaction efficiency.

## Proposal
A 6 byte address ID system, whereby all addresses used in Fuel must be registered either in the block they are used, or a previous block.

## Schema
The ID specifier is made up of two parts concatenated together, a (1) 4 byte block number and a (2) two byte address registration index.

Each block header will commit to a (1) Merkle root of all addresses registered, (2) an array of addresses being registered (i.e. from 0 - N), and (3) a commitment hash of the addresses.

### Breakdown:
```js
32 bits [Block Number]
16 bits [Index/Bitmap ID] 
```

### Block Header Example
```solidity
struct BlockHeader {
     ... block header data ...
     bytes32[] addresses;
     bytes32 address_commitment;
     bytes32 address_root;
}
```

## Example
```js
// This would reduce to address number 5 registered in block number 3.
0x000000030005
```

## Proofing
When doing any kind of address based fraud or transaction proofing, the addresses used in the proof (e.g. for signature or witness verification), will require the addition of a Merkle proof and block block header.

This means, for each address in question that is required to provide proofs, an SSLOAD and Merkle proof reduction is required. However, this is not a horrible expense as it is only required for fraud proofing, and should only be required for witness checking and witness verification.

## Invalid Merkle Root Submission
If an invalid Merkle root is submitted for the BlockHeader, this can be easily checked by building a merkle tree out of the address for that block and checking the root against the address root.

## Implementation Notes
- Each block could register their Merkle root in a separate SSTORE mapping on Ethereum instead of just the block commitments, which means that when doing the address proofing, you wont need to provide the full block header per address, just the Merkle proof (which would reduce to) the registered Merkle root.

Example:
```solidity
// block number => address merkle root
mapping(uint32 => bytes32) public address_roots;
```

- Alternatively, block headers can be provided for each address when in use in proofing.

### Example:
```solidity
struct AddressPoof {
    bytes32 address;
    bytes32[] merkleProof;
    BlockHeader header;
}
```

## Changes to Transaction Format

This would mean that instead of using 32 byte addresses in the transaction format, we would only need to include the 6 byte identifier. This would alter the transaction specification.

[...]

PR #443, opened 2022-12-09
Make message retryable and not spendable on gas

pixelcircuits:
Input messages are retryable until included in a script transaction that doesn't revert. The `amount` on an input message can no longer be spent on gas. Messages with no data are now spendable as an input coin (and only spendable as an input coin).

### DIFF
--- a/src/protocol/tx_format/input.md
+++ b/src/protocol/tx_format/input.md
@@ -19,20 +19,21 @@ Transaction is invalid if:
 
 ## InputCoin
 
-| name                  | type                         | description                                                            |
-|-----------------------|------------------------------|------------------------------------------------------------------------|
-| `txID`                | `byte[32]`                   | Hash of transaction.                                                   |
-| `outputIndex`         | `uint8`                      | Index of transaction output.                                           |
-| `owner`               | `byte[32]`                   | Owning address or predicate root.                                      |
-| `amount`              | `uint64`                     | Amount of coins.                                                       |
-| `asset_id`            | `byte[32]`                   | Asset ID of the coins.                                                 |
-| `txPointer`           | [TXPointer](./tx_pointer.md) | Points to the TX whose output is being spent.                          |
-| `witnessIndex`        | `uint8`                      | Index of witness that authorizes spending the coin.                    |
-| `maturity`            | `uint32`                     | UTXO being spent must have been created at least this many blocks ago. |
-| `predicateLength`     | `uint16`                     | Length of predicate, in instructions.                                  |
-| `predicateDataLength` | `uint16`                     | Length of predicate input data, in bytes.                              |
-| `predicate`           | `byte[]`                     | Predicate bytecode.                                                    |
-| `predicateData`       | `byte[]`                     | Predicate input data (parameters).                                     |
+| name                  | type                                   | description                                                            |
+|-----------------------|----------------------------------------|------------------------------------------------------------------------|
+| `txID`                | `byte[32]`                             | Hash of transaction or ID of originating message ID.                   |
+| `outputIndex`         | `uint8`                                | Index of transaction output.                                           |
+| `owner`               | `byte[32]`                             | Owning address or predicate root.                                      |
+| `amount`              | `uint64`                               | Amount of coins.                                                       |
+| `asset_id`            | `byte[32]`                             | Asset ID of the coins.                                                 |
+| `txPointer`           | [TXPointer](./tx_pointer.md)           | Points to the TX whose output is being spent.                          |
+| `messagePointer`      | [MessagePointer](./message_pointer.md) | Points to the message being spent.                                     |
+| `witnessIndex`        | `uint8`                                | Index of witness that authorizes spending the coin.                    |
+| `maturity`            | `uint32`                               | UTXO being spent must have been created at least this many blocks ago. |
+| `predicateLength`     | `uint16`                               | Length of predicate, in instructions.                                  |
+| `predicateDataLength` | `uint16`                               | Length of predicate input data, in bytes.                              |
+| `predicate`           | `byte[]`                               | Predicate bytecode.                                                    |
+| `predicateData`       | `byte[]`                               | Predicate input data (parameters).                                     |
 
 Given helper `len()` that returns the number of bytes of a field.
 
@@ -47,11 +48,13 @@ Transaction is invalid if:
 
 If `h` is the block height the UTXO being spent was created, transaction is invalid if `blockheight() < h + maturity`.
 
-> **Note:** when signing a transaction, `txPointer` is set to zero.
+> **Note:** when signing a transaction, `txPointer` and `messagePointer` are set to zero.
 >
-> **Note:** when verifying a predicate, `txPointer` is initialized to zero.
+> **Note:** when verifying a predicate, `txPointer` and `messagePointer` are initialized to zero.
 >
-> **Note:** when executing a script, `txPointer` is initialized to zero.
+> **Note:** when executing a script, `txPointer` and `messagePointer` are initialized to zero.
+>
+> **Note:** a message from the base chain with data length of zero is spent as a coin input using the message ID as the `txID` and `outputIndex` set to zero.
 
 The predicate root is computed identically to the contract root, used to compute the contract ID, [here](../id/contract.md).
 
@@ -78,26 +81,28 @@ Transaction is invalid if:
 
 ## InputMessage
 
-| name                  | type       | description                                                  |
-|-----------------------|------------|--------------------------------------------------------------|
-| `messageID`           | `byte[32]` | The messageID as described [here](../id/utxo.md#message-id). |
-| `sender`              | `byte[32]` | The address of the message sender.                           |
-| `recipient`           | `byte[32]` | The address or predicate root of the message recipient.      |
-| `amount`              | `uint64`   | Amount of base asset coins sent with message.                |
-| `nonce`               | `uint64`   | The message nonce.                                           |
-| `witnessIndex`        | `uint8`    | Index of witness that authorizes spending the coin.          |
-| `dataLength`          | `uint16`   | Length of message data, in bytes.                            |
-| `predicateLength`     | `uint16`   | Length of predicate, in instructions.                        |
-| `predicateDataLength` | `uint16`   | Length of predicate input data, in bytes.                    |
-| `data`                | `byte[]`   | The message data.                                            |
-| `predicate`           | `byte[]`   | Predicate bytecode.                                          |
-| `predicateData`       | `byte[]`   | Predicate input data (parameters).                           |
+| name                  | type                                   | description                                                  |
+|-----------------------|----------------------------------------|--------------------------------------------------------------|
+| `messageID`           | `byte[32]`                             | The messageID as described [here](../id/utxo.md#message-id). |
+| `sender`              | `byte[32]`                             | The address of the message sender.                           |
+| `recipient`           | `byte[32]`                             | The address or predicate root of the message recipient.      |
+| `amount`              | `uint64`                               | Amount of base asset coins sent with message.                |
+| `nonce`               | `uint64`                               | The message nonce.                                           |
+| `messagePointer`      | [MessagePointer](./message_pointer.md) | Points to the message being spent.                           |
+| `witnessIndex`        | `uint8`                                | Index of witness that authorizes spending the coin.          |
+| `dataLength`          | `uint16`                               | Length of message data, in bytes.                            |
+| `predicateLength`     | `uint16`                               | Length of predicate, in instructions.                        |
+| `predicateDataLength` | `uint16`                               | Length of predicate input data, in bytes.                    |
+| `data`                | `byte[]`                               | The message data.                                            |
+| `predicate`           | `byte[]`                               | Predicate bytecode.                                          |
+| `predicateData`       | `byte[]`                               | Predicate input data (parameters).                           |
 
 Given helper `len()` that returns the number of bytes of a field.
 
 Transaction is invalid if:
 
 - `witnessIndex >= tx.witnessesCount`
+- `dataLength == 0`
 - `dataLength > MAX_MESSAGE_DATA_LENGTH`
 - `predicateLength > MAX_PREDICATE_LENGTH`
 - `predicateDataLength > MAX_PREDICATE_DATA_LENGTH`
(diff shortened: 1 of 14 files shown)


pixelcircuits:
This PR is still very much in a draft state. It currently tries to redefine transaction validity to exclude message amounts being spent on gas (which I'm now realizing we could put this requirement in a message's predicate instead of changing the client validating rules). 

The ugliest part (IMO) is how it currently tries to keep track of which `InputMessages` have been "spent". There's also currently a hole in the design where a messages amount would get swept into a changeoutput during a failed transaction.

src/protocol/tx_validity.md
```diff
@@ -83,39 +85,49 @@ If this check passes, the UTXO ID `(txID, outputIndex)` fields of each contract
 For each asset ID `asset_id` in the input and output set:
 
 ```py
+def sum_messages(tx, asset_id) -> int:
```
Braqzen, line 88: You could also provide `type hints` for the parameters

src/protocol/tx_format/input.md
```diff
@@ -19,20 +19,21 @@ Transaction is invalid if:
...
-| `maturity`            | `uint32`                     | UTXO being spent must have been created at least this many blocks ago. |
-| `predicateLength`     | `uint16`                     | Length of predicate, in instructions.                                  |
-| `predicateDataLength` | `uint16`                     | Length of predicate input data, in bytes.                              |
-| `predicate`           | `byte[]`                     | Predicate bytecode.                                                    |
-| `predicateData`       | `byte[]`                     | Predicate input data (parameters).                                     |
+| name                  | type                                   | description                                                            |
+|-----------------------|----------------------------------------|------------------------------------------------------------------------|
+| `txID`                | `byte[32]`                             | Hash of transaction or ID of originating message ID.                   |
```
nfurfaro, line 24: So in the case of a retried message, this is the tx id from the initial attempt to spend/use the message?

nfurfaro (commented):
Generally looks good.

src/protocol/tx_format/message_pointer.md
```diff
@@ -0,0 +1,9 @@
+# MessagePointer
+
+The location of the message on the base chain. It can be used by UTXOs as a reference to the message or by the transaction itself to make it unique.
+
+| name          | type     | description        |
+|---------------|----------|--------------------|
+| `blockHeight` | `uint32` | Block height.      |
+| `txIndex`     | `uint16` | Transaction index. |
+| `logIndex`    | `uint16` | Log index.         |
```
Voxelot, line 9: This is fairly EVM specific, is this really necessary if we already have a nonce?

src/protocol/tx_format/input.md
pixelcircuits: In this scenario (message had no data and just has an amount) there isn't really a concept of retry. The amount would just move along to whatever outputs were specified. A retried message is just the same `InputMessage` until it is successfully spent.

src/protocol/tx_format/message_pointer.md
pixelcircuits: It needs to serve the same purpose as the txPointer, which I assume is for helping in transaction validation when a block is published? If that's the case, shouldn't it be EVM specific? The nonce doesn't currently help all that much because `nonce` is not indexed on the Ethereum event. We could make it indexed, but I also don't think we should be assuming Ethereum logs will never be pruned from full Ethereum nodes. I know there are several EIPs on the roadmap for log pruning to help ease state growth concerns

Voxelot: I see, is it right to assume we don't store any message metadata on the bridging contract itself and it just does the log emitting?

Would it be viable to just reuse the nonce field as a concatenation of these fields to lighten our tx payload size?

pixelcircuits: Correct, nothing is stored only an event is emitted to save gas. Using the nonce wouldn't lessen the payload size since it's a `uint64` (vs a `uint32` + `uint16` + `uint16`)

Voxelot: Would we still need the nonce on L2 if we have the pointer?

pixelcircuits: You mean remove the nonce from `InputMessage`? I guess it's not necessary since it's implicitly part of the messageId and it's not like it helps the client query Ethereum for the message in question. 

Side note: is there a way for newly syncing nodes to quickly download and trust indexed data about messages from other nodes? Is there a way to do this for the UTXO set in order to quickly catch up? I'm not sure how this could be done since our header doesn't include any state items like that.

Voxelot: > You mean remove the nonce from InputMessage?

Yep, as it seems like the message pointer already serves as a nonce. No need to have two nonces.

> is there a way for newly syncing nodes to quickly download and trust indexed data about messages from other nodes?

This is something we could enable a toggle option for, but it means trusting the sequencer. Potentially useful for infura type services but I wouldn't advise this as a default when running a full-node for any high-risk off-chain bridging (e.g. centralized custodial services).

Voxelot: closing this thread. I think keeping the nonce is fine as replacing it with message pointer may not be as straightforward and would be a larger breaking change (e.g. message id and signing procedures would need to have breaking changes)

src/protocol/tx_format/input.md
```diff
@@ -19,20 +19,21 @@ Transaction is invalid if:
...
+| `outputIndex`         | `uint8`                                | Index of transaction output.                                           |
+| `owner`               | `byte[32]`                             | Owning address or predicate root.                                      |
+| `amount`              | `uint64`                               | Amount of coins.                                                       |
+| `asset_id`            | `byte[32]`                             | Asset ID of the coins.                                                 |
+| `txPointer`           | [TXPointer](./tx_pointer.md)           | Points to the TX whose output is being spent.                          |
+| `messagePointer`      | [MessagePointer](./message_pointer.md) | Points to the message being spent.                                     |
```
Voxelot, line 30: I don't think adding message fields to coins is a clean solution, as fields like tx and outputIndex are no longer relevant. We also lose track of the message ID which is important for verifying in the consensus rules.

I'd prefer to have a new distinct type representing Coins that come from messages, which don't have conflicting identifier fields (e.g. `MessageCoin` or `DepositCoin`)

Voxelot:
Updated this design to keep messages separate from coins, while still allowing message with non-zero dataLength to be retryable

xgreenx:
Honestly, I don't understand why we need `MessagePointer`. Maybe someone can explain use cases for me=) Because it doesn't participate in the `MessageId` calculation, so we don't need it for uniqueness. How can it be used in fraud proofs? Right now, we have only `MessageRoot` in the block, so we will use the `MessageId` to do checks, not `MessagePointer`(we don't include messages from the relayer into this root, but maybe we should).

The same question I have for `TxPointer`. Do we really use it? I tried to find the reason for it, but it seems it is unclear from the commit history. We added `TXOPointer` in https://github.com/FuelLabs/fuel-specs/pull/249. But it seems the reason was that we already added it for the compressed transaction format in https://github.com/FuelLabs/fuel-specs/pull/135. And it seems the primary motivation for the compressed transaction format was the issue https://github.com/FuelLabs/fuel-specs/issues/111 to make the verification cheaper on the L1 side. But is that issue still actual? Because we still have `UtxoId` in the `InputCoin` and this `TxPointer`, so instead of 32 or 7/6 bytes, we have 32 + 7/6=)

It seems to me like we decided to optimize the verification on the L1 side with the usage of other primitives, but we stopped at some point and kept both variants. Also, maybe we can improve something in the design of fraud proofs after our last updates regarding bridging.

I'm not against the `MessagePointer` and `TxPointer`. I only try to understand why we simultaneously need (`MessagePointer` and `MessageId`) or (`UtxoId` and `TxPointer`).

pixelcircuits:
My understanding is that the "pointers" are supposed to make the fault proofs easier (potentially single step fault proofs). It lets someone directly challenge where the proposed block is claiming its data is from (as opposed to having to challenge for this data from the proposer first). However, even if we can prove fault in a single step in this case, proving fault in VM execution will almost certainly have to be proposer/challenger interactive. So you're right, I'm not sure what we're really gaining... 

IMO, we should be optimizing for the optimistic case where faults proofs will never even be initiated. So I'd vote we eventually remove the pointers altogether and save on data. Embracing the optimistic case at the cost of increased fault proof complexity seems to be the winning OR design currently.

`MessagePointer` was added for several reasons. The first being to just put it on par with how InputCoins have a pointer. The second was to help facilitate fault proofs since the `messageId` is something the contracts don't know about and is actually computed off-chain. Again, we could just add an interactive fault proof step where the proposer is challenged for this data (technically anyone can respond with the correct data on behalf of the proposer).

I'd vote we drop the `MessagePointer` on this spec change and eventually drop `TxPointer` on InputCoins too.

xgreenx:
I also vote for removing it if we can achieve all functionality that we need without it=) Do we need to wait for comments from @Voxelot and @adlerjohn before doing that?

src/protocol/tx_format/input.md
```diff
@@ -78,20 +78,21 @@ Transaction is invalid if:
...
-| `witnessIndex`        | `uint8`    | Index of witness that authorizes spending the coin.          |
-| `dataLength`          | `uint16`   | Length of message data, in bytes.                            |
-| `predicateLength`     | `uint16`   | Length of predicate, in instructions.                        |
-| `predicateDataLength` | `uint16`   | Length of predicate input data, in bytes.                    |
-| `data`                | `byte[]`   | The message data.                                            |
-| `predicate`           | `byte[]`   | Predicate bytecode.                                          |
-| `predicateData`       | `byte[]`   | Predicate input data (parameters).                           |
+| name                  | type                                   | description                                                  |
+|-----------------------|----------------------------------------|--------------------------------------------------------------|
+| `messageID`           | `byte[32]`                             | The messageID as described [here](../id/utxo.md#message-id). |
```
xgreenx, line 83: I know it was here before, but why must we duplicate the `message_id` field if we can compute it based on the other fields?

image

I have only one assumption: because we want to put it into the memory of the `vm` to be able to read it, instead of each time calculation.

But does this optimization worth 32 bytes per input?

@adlerjohn @Voxelot

Voxelot: Hmm it does seem like we could omit the message id on inputs, since we have to re-compute the hash to validate it before spending anyways.

I'm also not sure how fraud-proofs would verify the messages. Since we can't lookup old events / logs, unless we put the message id into storage I'm not sure which fields are actually needed. cc @pixelcircuits

pixelcircuits: So, we may end up putting the messageIds into storage for simplicity, but it's actually possible to prove emitted events given a starting block hash (you navigate down the chain and provide the necessary merkle proofs and hash preimages as necessary to show the log that occurred). I know Optimism intends to use this approach with Bedrock (essential complicate the fault proving, but optimizing for the optimistic case). This can also potentially be done in a zkProof via something like [Herodotus](https://www.herodotus.dev/) (their main focus is to provide proofs of other chain history, but it could be used to prove Ethereum history on Ethereum itself). It would actually be really cool to use zkProofs like Herodotus to prove output messages in our Fuel chain instead of having a list of output messages always copied into the block header

pixelcircuits:
> I know it was here before, but why must we duplicate the message_id field if we can compute it based on the other fields?

I think I agree. It's kinda funny because earlier in the comments of this PR, the idea of dropping the `nonce` on `InputMessage` was being thrown around. This suggestion is similar but we drop the `messageId` instead and use the `nonce` as more of the raw identifier. 

I'm also starting to think that we should update `MessageOutput` too. Right now it just has an `amount` and `recipient` field. The `amount` field is there so that base asset can essentially be burned by the protocol before being withdrawn from the bridge. The `recipient` field was added because we needed something about the output to get filled so it can only be `SMO`'d once (since a message with an amount of 0 is valid). I'm thinking we should replace `recipient` with `nonce` since that's probably more useful (although it's a bit odd to have it zeroed out because it is technically knowable at time of signing).

I'm also confused about where the actual output message data is getting stored/logged. It's not on the `OutputMessage` and it's not on the `MessageOut` receipt. I'm also not sure how these 'OutputMessage`s could survive a regenesis since they're not part of the initial json file.

Voxelot:
>  I'm thinking we should replace recipient with nonce since that's probably more useful (although it's a bit odd to have it zeroed out because it is technically knowable at time of signing).

This also seems redundant, as it's storing data that should be derived from the existence of the output on the tx. If all we need is an output for the purpose of burning coins and all the data really comes from the receipt anyways, then we maybe we could reduce the output to only contain an amount? That way we can easily verify sum of inputs >= sum of outputs, and we don't include data unrelated to block validity? Or do we need some malleable field that can completely commit to the receipt for fraud proving?

pixelcircuits:
If we only have `amount`, then we can probably tweak the SMO opcode to not even need to touch an `OutputMessage` if the amount is 0. Are there any problems with this? We would have a `MessageOut` receipt with no corresponding `OutputMessage`.

Also, where is the message data stored? The receipt only has a hash of the data. It feels kinda weird for the message data to not be returned after executing a transaction and instead we have to make a separate graphQL query to actually get it. But maybe not since a user will always have to make a separate graphQL to get the message proof anyway.

xgreenx:
Actually, I'm not sure why we need `OutputMessage` because we can't spend it on the `Fuel` chain.

image

And we don't process it in the `fuel-core`, right now. Maybe we should start to check that receipts contain at least the same count of `MessageOut` as the count of `MessageOutput` and that fields are the same(but for fields, it is partially useless because the user didn't sign them, and it is a double-check of the `fuel-vm` execution).

image

When we added this output, it participated in the `sum_output` calculation: 

image

But now, we don't do that(because the user does not sign it, I think), and this output seems partially useless to me. It only helps on the UI side to show that transaction has withdrawals.

image

It may be worth considering removing this output. @adlerjohn

src/protocol/tx_format/input.md
xgreenx: @pixelcircuits  So, what is final conclusion?=) Do we need to keep the message id in the transaction, or it is okay to compute it during fraud proofs because we have access to all fields?

Voxelot (approved):

Voxelot (approved):

adlerjohn (approved):

Issue #435, opened 2022-11-28
Add `messages` and their types to the JSON ABI


mohammadfawaz:
This would be similar to logs. We could have a separate array called `messagesTypes` that lists all the types used, just like `loggedTypes`. We can then encode the type IDs as the first word of each message because we don't have a free register in `smo` that we can use.

PR #444, opened 2022-12-09
Add `messagesTypes` to the JSON ABI

mohammadfawaz:
Closes https://github.com/FuelLabs/fuel-specs/issues/435

This is basically identical to `loggedTypes` except that, instead of having a special register on the instruction to hold the ID, the message data itself holds the ID as the fist word in it. For example, if the message sent is a `1u64`, and the ID of the type declaration `u64` in the JSON ABI is `3`, then message data would look like:
```rust
0x00000003_00000001
```


ra0x3 (approved):
LGTM but will defer to others. Thanks @mohammadfawaz

PR #445, opened 2022-12-13
fixed improper range description on SRWQ

mitch-fuel:
SRWQ Description stated that state retrieval would occur for a range of keys specified starting at a single location in memory, instead of a range from state extending from a single specified key.


mohammadfawaz (approved):
Looks good. Thanks!

Voxelot (approved):

PR #448, opened 2023-01-17
Changes for predicate gas metering

mitch-fuel:


src/vm/index.md
```diff
@@ -105,7 +105,7 @@ For any input of type [`InputType.Coin`](../protocol/tx_format/index.md), a non-
 
 For each such input in the transaction, the VM is [initialized](#vm-initialization), then:
 
-1. `$pc`  and `$is` are set to the start of the input's `predicate` field.
+1. `$pc` and `$is` are set to the start of the input's `predicate` field.
```
Voxelot, line 108: Maybe we should be more specific here around how gas factors into predicate verification? If we introduce the concept of `remaining_gas` here it would make the script execution update more clear.

mitch-fuel: how should we ensure that we don't overly specify implementation for potentially parallelizing predicate execution? we eventually wouldn't be initializing predicate execution with remaining gas each time, just aggregating the resultant gas used.

Voxelot: Good point, although even if we don't include remaining gas in initialization it should still be explicitly defined before using it for script initialization. Something like:

```math
remaining_gas =  \sum_{i=0}^p gas(predicates[i])
```

src/vm/index.md
```diff
@@ -123,7 +130,7 @@ If script bytecode is present, transaction validation requires execution.
 The VM is [initialized](#vm-initialization), then:
 
 1. `$pc` and `$is` are set to the start of the transaction's script bytecode.
-1. `$ggas` and `$cgas` are set to `tx.gasLimit`.
+1. `$ggas` and `$cgas` are set to `tx.gasLimit`. If predicate execution has decreased remaining gas, gas limit is initialized to remaining gas.
```
Voxelot, line 133:
```suggestion
1. `$ggas` and `$cgas` are set to lower of `tx.gasLimit` or `remaining_gas`. `remaining_gas` is determined by deducting any gas consumed during predicate execution from `tx.gasLimit`.
```

src/vm/index.md
```diff
@@ -105,7 +105,14 @@ For any input of type [`InputType.Coin`](../protocol/tx_format/index.md), a non-
 
 For each such input in the transaction, the VM is [initialized](#vm-initialization), then:
 
-1. `$pc`  and `$is` are set to the start of the input's `predicate` field.
+1. `$pc` and `$is` are set to the start of the input's `predicate` field.
+1. `$ggas` and `$cgas` are set to the lower of `tx.gasLimit` or the remaining gas following the previous predicate execution.
```
adlerjohn, line 109: Hmmm. There's a few issues with this:

1. It creates a sequential dependence on lower-position predicates. Without any prior information on how much gas is used by each predicate, this requires predicates be verified sequentially.
2. Even _with_ prior information, this change makes predicates no longer position-independent. Which can make partially-signed transactions more complex to implement.

Overall, I'd prefer something like a fixed max gas for each predicate, independently. And this max gas can be a consensus rule.

If the gas schedule is changed, then predicates may become invalid. But that's the case even if we allow users to define the max gas since there's a hard cap in the form of the block gas limit.

adlerjohn (changes requested):

src/vm/index.md
Voxelot: Good points. Having a per predicate gas limit would be better for parallelization. Would there be both a per-predicate gasLimit specified + consensus rule max gas per predicate? Or would each predicate always assume it has the consensus based max gas, potentially locking up more funds than needed for execution?

The main benefits of using the tx-level gas limit as opposed to predicate level is:
1. UX simplicity - it is easier for the sdk to estimate the tx gas limit regardless of changes to the gas schedule
2. Smaller tx payload, as each new field increases our calldata expense / bandwidth
3. Users only need to lock up the minimal amount of necessary funds to pay for execution, potentially allowing for more user-side parallelism by minimizing gas lockup overhead (assuming there's no per-predicate user configurable gas limit)

Some ways to do parallel execution with this approach could be:

- Simply execute all predicates in parallel using the same global tx.gasLimit for each one. Total up the amount of gas each predicate consumed and fail the verification after execution if they ran over the limit. In the best case this doesn't add any overhead and keeps the predicate UX simpler. In the worst case the VM may spend more time executing instructions than it should.
   - The main risk with this approach is if the txpool would do extra work, it's possible it could be overloaded more easily. However, since it has to evaluate potentially invalid predicates anyways without being able to collect any fees, it's hard to say if this really worsens the situation for the txpool. 
   - The other risk would be during bft consensus. If a block proposed for the next round exceeded the allowed execution budget it could cause a lag in voting (although it would have to be some very large predicates at the very end of the block). However, given BFT only supports a small validator set, proposing a bad block which exceeds the execution budget could be a punishable/slashable offence which would deter this from happening.
- Use an atomic counter external to the cgas/ggas registers to track the total gas usage across multiple VM instances. However, this could add execution overhead as charging gas for predicates would require an additional CAS operation.

Voxelot: We had a call with @FuelLabs/client and @adlerjohn to discuss this in more detail -  summarizing the notes here:

- Sharing a global gasLimit prevents parallelization between predicates and script execution. We need to know how much gas predicates can use in advance, otherwise if there's an overage we would lose determinism as to whether the script status is reverted or if the predicates are invalid, thus invalidating the entire block.
- We will add a new malleable field to predicates called gasUsed. This avoids the need to have predicate amounts cover a global gas limit per predicate which could create a lot of dust.
- Each predicate will have cgas and ggas always initialized to the tx.gasLimit
- If predicate.gasUsed is set, the vm will ensure the current gas consumption doesn't exceed that (resolving potential DOS concerns)
- predicate.gasUsed will not be signed over, but the mempool will expect it to be set correctly by the user
- The user should be able to estimate the gas usage of each predicate via a single tx dry-run
- predicate.gasUsed is zeroed out on the tx during predicate verification

xgreenx: > The user should be able to estimate the gas usage of each predicate via a single tx dry-run

And we can execute the predicate without a `dry-run` on the SDK side potentially in the future(because we don't need access to the on-chain state)

src/vm/index.md
```diff
@@ -94,20 +95,42 @@ Then the following registers are initialized (without explicit initialization, a
...
 
 For any input of type [`InputType.Coin`](../protocol/tx_format/index.md), a non-zero `predicateLength` field means the UTXO being spent is a [P2SH](https://en.bitcoinwiki.org/wiki/P2SH) rather than a [P2PKH](https://en.bitcoinwiki.org/wiki/Pay-to-Pubkey_Hash) output.
 
 For each such input in the transaction, the VM is [initialized](#vm-initialization), then:
 
-1. `$pc`  and `$is` are set to the start of the input's `predicate` field.
+1. `$pc` and `$is` are set to the start of the input's `predicate` field.
+1. `$ggas` and `$cgas` are set to `tx.gasLimit`.
+
+Predicate verification will fail if gas is exhausted during execution. Additionally, if `predicate.gasUsed` is nonzero, a check will be performed to ensure that less than `predicate.gasUsed` gas is consumed:
```
Voxelot, line 127: We shouldn't include the non-zero logic here now that we have estimation.

```suggestion
Predicate verification will fail if gas is exhausted during execution. Additionally, a check will be performed to ensure that less than `predicate.gasUsed` gas is consumed:
```

src/vm/index.md
```diff
@@ -94,20 +95,42 @@ Then the following registers are initialized (without explicit initialization, a
...
+
+```pseudo
+if $cgas < tx.gasLimit - predicate.gasUsed {
+    return false
+}
```
Voxelot, line 132:
```suggestion
# invert predicate.gasUsed by the tx.gasLimit to simplify comparison with $cgas
min_allowed_gas = tx.gasLimit - predicate.gasUsed
if $cgas < min_allowed_gas {
    return false
}
```

src/vm/index.md
```diff
@@ -118,14 +141,16 @@ In addition, during predicate mode if `$pc` is set to a value greater than the e
 
 A predicate that halts without returning Boolean `true` does not pass verification, making the entire transaction invalid. Note that predicate validity is monotonic with respect to time (i.e. if a predicate evaluates to `true` then it will always evaluate to `true` in the future).
```
Voxelot, line 142: > During predicate mode, hitting any of the following instructions causes predicate verification to halt, returning Boolean `false`:
> 1. Any [contract instruction](./instruction_set.md#contract-instructions).
> 1. [JMP](./instruction_set.md#jmp-jump), [JI](./instruction_set.md#ji-jump-immediate), [JNE](./instruction_set.md#jne-jump-if-not-equal), [JNEI](./instruction_set.md#jnei-jump-if-not-equal-immediate), or [JNZI](./instruction_set.md#jnzi-jump-if-not-zero-immediate) with jump-to value less than or equal to `$pc` (these would allow loops). In other words, `$pc` must be strictly increasing.

The section above this needs to adjust the restriction on JMP to allow backwards jumps within the predicate range

src/vm/index.md
```diff
@@ -94,20 +95,42 @@ Then the following registers are initialized (without explicit initialization, a
...
+1. `$pc` and `$is` are set to the start of the input's `predicate` field.
+1. `$ggas` and `$cgas` are set to `tx.gasLimit`.
+
+Predicate verification will fail if gas is exhausted during execution. Additionally, if `predicate.gasUsed` is nonzero, a check will be performed to ensure that less than `predicate.gasUsed` gas is consumed:
+
+```pseudo
+if $cgas < tx.gasLimit - predicate.gasUsed {
+    return false
+}
+```
```
Voxelot, line 133: We should also have a check that the sum of all `predicate.gasUsed` <= `tx.gasLimit`

src/protocol/tx_format/input.md
```diff
@@ -105,5 +108,6 @@ Transaction is invalid if:
 - `dataLength != len(data)`
 - `predicateLength * 4 != len(predicate)`
 - `predicateDataLength != len(predicateData)`
+- `predicateGasUsed > MAX_GAS_PER_PREDICATE`
```
Voxelot, line 111: We should also have a check that total sum of `predicate.gasUsed` <= `tx.gasLimit`.

src/vm/index.md
```diff
@@ -94,20 +95,42 @@ Then the following registers are initialized (without explicit initialization, a
...
+For any input of type [`InputType.Coin`](../protocol/tx_format/index.md), a non-zero `predicateLength` field means the UTXO being spent is a [P2SH](https://en.bitcoinwiki.org/wiki/P2SH) rather than a [P2PKH](https://en.bitcoinwiki.org/wiki/Pay-to-Pubkey_Hash) output.
+
+For each such input in the transaction, the VM is [initialized](#vm-initialization), then:
+
+1. `$pc` and `$is` are set to the start of the input's `predicate` field.
+1. `$ggas` and `$cgas` are set to `tx.gasLimit`.
+
+Predicate verification will fail if gas is exhausted during execution.
+
+After successful execution, `predicate.gasUsed` is set to `tx.gasLimit - $cgas`.
```
Voxelot, line 116: Even though we don't allow calls, I think it's safer to use ggas instead of cgas.

```suggestion
After successful execution, `predicate.gasUsed` is set to `tx.gasLimit - $ggas`.
```

src/vm/instruction_set.md
```diff
@@ -1784,81 +1784,83 @@ Set `$rA` to the index of the currently-verifying predicate.
...
+| `GTF_INPUT_COIN_AMOUNT`                    | `0x105` | `tx.inputs[$rB].amount`                             |
+| `GTF_INPUT_COIN_ASSET_ID`                  | `0x106` | Memory address of `tx.inputs[$rB].asset_id`         |
+| `GTF_INPUT_COIN_TX_POINTER`                | `0x107` | Memory address of `tx.inputs[$rB].txPointer`        |
+| `GTF_INPUT_COIN_WITNESS_INDEX`             | `0x108` | `tx.inputs[$rB].witnessIndex`                       |
+| `GTF_INPUT_COIN_MATURITY`                  | `0x109` | `tx.inputs[$rB].maturity`                           |
+| `GTF_INPUT_COIN_PREDICATE_LENGTH`          | `0x10A` | `tx.inputs[$rB].predicateLength`                    |
+| `GTF_INPUT_COIN_PREDICATE_DATA_LENGTH`     | `0x10B` | `tx.inputs[$rB].predicateDataLength`                |
+| `GTF_INPUT_COIN_PREDICATE`                 | `0x10C` | Memory address of `tx.inputs[$rB].predicate`        |
+| `GTF_INPUT_COIN_PREDICATE_DATA`            | `0x10D` | Memory address of `tx.inputs[$rB].predicateData`    |
+| `GTF_INPUT_COIN_PREDICATE_GAS_USED`        | `0xXXX` | `tx.inputs[$rB].predicate.gasUsed`                  |
```
Voxelot, line 1831:
```suggestion
```

We may not need GTF on this field since it is malleable and might change between predicate estimation and verification.

src/vm/instruction_set.md
```diff
@@ -1784,81 +1784,83 @@ Set `$rA` to the index of the currently-verifying predicate.
...
+| `GTF_INPUT_MESSAGE_AMOUNT`                 | `0x117` | `tx.inputs[$rB].amount`                             |
+| `GTF_INPUT_MESSAGE_NONCE`                  | `0x118` | `tx.inputs[$rB].nonce`                              |
+| `GTF_INPUT_MESSAGE_WITNESS_INDEX`          | `0x119` | `tx.inputs[$rB].witnessIndex`                       |
+| `GTF_INPUT_MESSAGE_DATA_LENGTH`            | `0x11A` | `tx.inputs[$rB].dataLength`                         |
+| `GTF_INPUT_MESSAGE_PREDICATE_LENGTH`       | `0x11B` | `tx.inputs[$rB].predicateLength`                    |
+| `GTF_INPUT_MESSAGE_PREDICATE_DATA_LENGTH`  | `0x11C` | `tx.inputs[$rB].predicateDataLength`                |
+| `GTF_INPUT_MESSAGE_DATA`                   | `0x11D` | Memory address of `tx.inputs[$rB].data`             |
+| `GTF_INPUT_MESSAGE_PREDICATE`              | `0x11E` | Memory address of `tx.inputs[$rB].predicate`        |
+| `GTF_INPUT_MESSAGE_PREDICATE_DATA`         | `0x11F` | Memory address of `tx.inputs[$rB].predicateData`    |
+| `GTF_INPUT_MESSAGE_PREDICATE_GAS_USED`     | `0xXXX` | `tx.inputs[$rB].predicate.gasUsed`                  |
```
Voxelot, line 1850:
```suggestion
```

We may not need GTF on this field since it is malleable and might change between predicate estimation and verification.

src/vm/instruction_set.md
adlerjohn: Hmmm. There are GTF for other malleable fields. It should either be consistently that such fields have a GTF, or not.

src/protocol/tx_format/input.md
mitch-fuel: this is now addressed in the transaction validity rules

src/protocol/tx_format/transaction.md
```diff
@@ -80,6 +80,7 @@ Transaction is invalid if:
 - `scriptDataLength > MAX_SCRIPT_DATA_LENGTH`
 - `scriptLength * 4 != len(script)`
 - `scriptDataLength != len(scriptData)`
+- `gasLimit` is less than the sum of all `predicateGasUsed` for `InputeType.Coin` or `InputType.Message` where predicate length is greater than zero.
```
Voxelot, line 83:
```suggestion
- `gasLimit` is less than the sum of all `predicateGasUsed` for `InputType.Coin` or `InputType.Message` where predicate length is greater than zero.
```

Voxelot (approved):

src/vm/index.md
```diff
@@ -94,38 +95,63 @@ Then the following registers are initialized (without explicit initialization, a
...
 
-1. `$pc`  and `$is` are set to the start of the input's `predicate` field.
+1. `$pc` and `$is` are set to the start of the input's `predicate` field.
+1. `$ggas` and `$cgas` are set to `predicateGasUsed`.
+
+Predicate verification will fail if gas is exhausted during execution.
 
 During predicate mode, hitting any of the following instructions causes predicate verification to halt, returning Boolean `false`:
 
 1. Any [contract instruction](./instruction_set.md#contract-instructions).
```
mohammadfawaz, line 139: Should this be reworded a bit since there is no list anymore - it's only one item. Maybe something like

> During predicate mode, hitting any [contract instruction](./instruction_set.md#contract-instructions) causes predicate verification to halt, returning Boolean `false`.

mitch-fuel: I agree, I'll get this in now.

mitch-fuel: done

Voxelot (approved):

xgreenx (approved):

PR #462, opened 2023-02-28
Chain Identifier

### LINKED ISSUE #340: Add chain ID to tx format

Voxelot:
Closes #340

Rather than adding chain ID as a new field to the transaction, this approach includes chain id as part of our identifier hashing procedures. This means predicates, contracts and transactions will be innately chain specific, preventing cross-chain replay.

Consider the scenario where there are two instances of fuel (e.g. app-specific rollups) that each have the same contracts deployed. If a user signs a transaction to transfer tokens without a chain identifier, the same transaction could potentially be replayed on the second chain unintentionally. While this would require the same input utxos to exist on both chains, there are still certain scenarios where this could occur and cause a single transaction to be used for multiple transfers across different chains.

I considered hashing `CHAIN_ID`'s into Message ID's, but didn't see any obvious benefits since it wouldn't be emitted by or committed to on L1. It may be worth considering an alternative approach such as a new chain_id field for messages to ensure the same message can't be used on multiple chains. However, since predicate messages are primarily used in the current bridging designs, this should already be covered for the majority of use-cases.


src/protocol/id/contract.md
```diff
@@ -1,3 +1,3 @@
...
+For a transaction of type `TransactionType.Create`, `tx`, the contract ID is `sha256(0x4655454C ++ CHAIN_ID ++ tx.data.salt ++ root(tx.data.witnesses[bytecodeWitnessIndex].data) ++ root_smt(tx.storageSlots))`, where `root` is the Merkle root of [the binary Merkle tree](../cryptographic_primitives.md#binary-merkle-tree) with each leaf being an 8-byte word of two instructions, and `root_smt` is the [Sparse Merkle tree](../cryptographic_primitives.md#sparse-merkle-tree) root of the provided key-value pairs. If the bytecode is not a multiple of 8 bytes (i.e. if there are an odd number of instructions), the last instruction is padded with 4-byte zero.
```
xgreenx, line 3: What are the benefits of adding chain id into the `ContractId` calculation? And what vulnerabilities do we have if we don't do that?

With the same `ContractId` is simpler for developers to deploy the several instances of the same project on different networks.

src/protocol/tx_format/constants.md
```diff
@@ -14,3 +14,4 @@
 | `MAX_MESSAGE_DATA_LENGTH`   | `uint16` |                 | Maximum length of message data, in bytes.     |
 | `MAX_STORAGE_SLOTS`         | `uint16` | `255`           | Maximum number of initial storage slots.      |
 | `MAX_WITNESSES`             | `uint64` | `16`            | Maximum number of witnesses.                  |
+| `CHAIN_ID`                  | `uint64` |                 | A unique per-chain identifier.                |
```
xgreenx, line 17: Do we want to calculate `CHAIN_ID` based on the genesis config, or it is a manually typed number?

src/protocol/id/contract.md
xgreenx: I see that it is important for the predicate to be unique per network because it allows using inputs. But I'm not sure that we need this rule for contracts.

xgreenx (commented):
Do we need to add a new opt code that allows getting chain id?

src/protocol/tx_format/constants.md
Voxelot: I think it would make the most sense as a genesis config parameter.

src/protocol/id/contract.md
Voxelot: True, we don't need chain specific contract ids to prevent replay attacks. The only benefit would be to ensure there is a globally unique id for contracts which could help with search & indexing of multiple chains.

Voxelot:
> Do we need to add a new opt code that allows getting chain id?

We could add it to the `GM` opcode as a new arg. Seems like it would be useful to have this.

src/protocol/id/contract.md
mitchmindtree: > The only benefit would be to ensure there is a globally unique id for contracts which could help with search & indexing of multiple chains.

Would this really be useful? If you're searching for a contract, you're probably searching for it on a particular chain. If anything, perhaps it could be useful for the contract ID to be universal, as you could more easily identify instances of the same contract on different networks?

It would be nice to have a good motivation to change the behaviour of contract IDs, as it does complicate some tooling features around `CONTRACT_ID` consts a little. That said, I guess we'll need to make a bunch of the same changes for predicate roots either way.

adlerjohn:
Chain ID would be useful not only for transactions, but also signed messages. E.g. https://eips.ethereum.org/EIPS/eip-712#definition-of-domainseparator

> uint256 chainId the [EIP-155](https://eips.ethereum.org/EIPS/eip-155) chain id. The user-agent should refuse signing if it does not match the currently active chain.

Voxelot:
Updated the PR:

- GM arg for CHAIN_ID
- Made contract ID chain agnostic
- Added separate predicate root ID which is chain specific.

xgreenx (approved):

mitchmindtree (approved):

PR #464, opened 2023-03-01
Make `$hp` point to immediately after usable memory.

### LINKED ISSUE #322: `$hp` is initialised to incorrect value.
At [VM initialization](https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/main.md#vm-initialization) it says:

> `$hp = VM_MAX_RAM - 1`: the heap area begins at the top and is empty to start.

It makes more sense for `$hp` to be initialised to just `VM_MAX_RAM`.  It always essentially points to either non-addressable memory or previously allocated buffers (via `ALOC`), both of which are essentially 'reserved'.

adlerjohn:
Fixes #322

### DIFF
--- a/src/vm/index.md
+++ b/src/vm/index.md
@@ -35,24 +35,24 @@ FuelVM instructions are exactly 32 bits (4 bytes) wide and comprise of a combina
 
 Of the 64 registers (6-bit register address space), the first `16` are reserved:
 
-| value  | register | name                | description                                                                   |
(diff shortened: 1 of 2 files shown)


mohammadfawaz:
Do we also need to update the description of `$hp` [here](https://github.com/FuelLabs/fuel-specs/blob/da8b36f18e256f92ec89db24b2483a25074fe4d4/src/vm/index.md?plain=1#L47)?

mohammadfawaz (approved):
Looks good!

PR #465, opened 2023-03-01
Add nested call limit

Dentosal:
See https://github.com/FuelLabs/fuel-vm/pull/370


Voxelot (approved):

adlerjohn:
How is this limit tracked in the VM? The current setup of the VM is that it's entirely fraud provable with only working memory and registers.

Voxelot:
Well all the callframes should be in memory right? So you could scan over those in a similar way to re-entrancy detection as part of the fraud-proof.

PR #469, opened 2023-03-13
Conditional jumps only check for address validity if the condition is true

Dentosal:


Dentosal:
> Is this for optimization purposes? Should we update the VM to always check the JMP args just for consistency?

It's faster to not check, but I'm not sure if this decision was done for optimization reasons. It's essentially saving us a couple of multiplications and maybe four local branches per each jump instruction where the branch is not selected. CPUs should be able to speculatively execute these anyway, so the performance penalty here is tiny.

However, I can imagine situations where not doing the check would allow the Sway compiler to produce smaller code. On the other hand, the possible gains are again minimal, since our instruction set is so inefficient anyway.

Still, one use case for not checking is combining a niche optimization with a jump table. Imagine that we have a counter that's `Option<u64>`, but we know that the max value will never be reached, so we reuse `u64::MAX` to encode None, so the whole value fits into a 8-byte register. Then when we want to do a jump table lookup using that value, we could do something like this:

```
/// Assume the $a register is of the above `Option<u64>` type
not $tmp, $zero                         // Set $tmp to u64::MAX
jnef $tmp, $a, $a, offset_to(.table)    // Do jump table jump only if we have Some(x)

// Here is the processing for None
rvrt

// This is the jump table
.table:
jmp foo
jmp bar
jmp baz
```

In this case, the check would fail since the jump target address is `u64::MAX` which is way above vm ram limit, but the jump is never performed with that value.

Voxelot (approved):

PR #472, opened 2023-03-17
Make `Nonce` as 32 bytes array

xgreenx:
The specs say that the `Nonce` is a 32 bytes array but the current code uses `Word`. In the follow-up PR in `fuel-vm` we will replace `Word` with a 32-byte array.

### DIFF
--- a/src/protocol/tx_format/input.md
+++ b/src/protocol/tx_format/input.md
@@ -82,20 +82,20 @@ Transaction is invalid if:
 
 ## InputMessage
 
-| name                  | type         | description                                             |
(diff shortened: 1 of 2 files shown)


Voxelot (dismissed):

Voxelot (approved):

Voxelot:
Also updated the fuel-tx serialized format to make input message nonce bytes32, since it can be larger than uint64 according to the message id specification which states that it's determined by the chosen L1.

pixelcircuits (approved):

nfurfaro (approved):

nfurfaro (approved):

PR #477, opened 2023-04-05
clarify the description of the coinbase instruction

Voxelot:
The previous description was somewhat ambiguous about whether the `CB` opcode returns the address for the block producer or it's coinbase address (which can be a distinct cold wallet for collecting fees).


mitch-fuel (approved):
LGTM

Issue #478, opened 2023-04-11
Big integer arithmetic


Voxelot:
FuelVM currently only support u64 based arithmetic. This is insufficient for most DEX usecases, which necessitates using sway libraries for larger integers. The problem with this approach is that due to in-lining and just the overall bulk associated with supporting operations like u256 division using only u64s, the contract sizes for even relatively simple contracts is growing very large. 

This can be partially solved through usage of LDC, however it also adds a lot of VM overhead if we can't take advantage of natively supported machine instructions for big int math. Since the EVM supports u256 by default, the fraud proof overhead of adding wide number arithmetic instructions should be minimal.

Research whether u128 opcodes would be sufficient to cut down most of the bytecode overhead, and which operations would be the most frequently used. E.g. A special `u128::mul_div` may increase performance a lot for DEX applications compared to using separate mul and div opcodes.

Dentosal:
See https://github.com/FuelLabs/fuel-core/issues/1117 for some initial discussion

Dentosal:
Done in https://github.com/FuelLabs/fuel-specs/pull/483

PR #483, opened 2023-04-13
Wideint instructions

Dentosal:
Progress towards https://github.com/FuelLabs/fuel-specs/issues/478.  Implementation PR: https://github.com/FuelLabs/fuel-vm/pull/424

This PR adds instructions for handling wideint types, specifically memory-stored 128- and 256-bit unsigned integers. The following opcodes are added:

* MLDV - Fused multiply-add of 64bit register values
* WDCM - Compare 128bit integers
* WQCM - Compare 256bit integers
* WDOP - Cheap 128bit operations (add,, sub, shr, shl, not, and, or, xor)
* WQOP - Cheap 256bit operations (add,, sub, shr, shl, not, and, or, xor)
* WDML - Multiply 128bit
* WQML - Multiply 256bit
* WDDV - Divide 128bit
* WQDV - Divide 256bit
* WDMD - Fused multiply-divide 128bit
* WQMD - Fused multiply-divide 256bit
* WDAM - AddMod 128bit
* WQAM - AddMod 256bit
* WDMM - MulMod 128bit
* WQMM - MulMod 256bit

The naming convention is to prefix the operations with their size: `WD` for double-word-sized (128 bit) operations, and `WQ` for quad-word-sized (256 bit) operations.

### Grouped by cost

The operations are grouped to these opcodes so that cost of a signle opcode can be a constant. To avoid polluting the opcode space, several operations are grouped together: W*CM for compare operations and W*OP for arithmetic operations. The actual operation is then selected using the immediate part of the instruction.

Since multiplication and especially division are more expensive than cheap operations behind these groups, they have their own, separate opcodes. Similarly AddMod and MulMod, which require larger precision than the input operations, are grouped separately. A separate modulus instruction is not provided, but AddMod with addition of zero will perform this function with minimal overhead.

### Direct/Indirect arguments

Some of the operations also allow users to select between direct and indirect operands. When a direct operand is selected, instead of reading the wide integer pointed by the register value, the register value itself is zero-extended to full-width value and used as an operand. This reduces the size of many common operations by a couple of instructions. Also, this means that u64 * u64 -> u128 can be done directly.

### What's missing?

Astute reviewers might have already noticed that some operations are missing. 

* exponentation - Can be added if needed, also ExpMod could be added
* logarithm - This can be added if any use cases are found
* nth root (or even square root) - The roots are rather slow to compute, and hopefully never needed. Even the 64bit version was tricky
* Fused opcodes like multiply-add and multiply-divide

### Hopes and dreams

We could support these operations for 64-bit integers as well, and remove current ALU instructions completely. However, I think that we are too committed to backwards compatibility of the current ops even now.

### DIFF
--- a/src/vm/instruction_set.md
+++ b/src/vm/instruction_set.md
@@ -688,7 +688,7 @@ Panic if:
 |             |                                                                                       |
 |-------------|---------------------------------------------------------------------------------------|
 | Description | Compare or examine two 128-bit integers using selected mode                           |
-| Operation   | `b = mem[$rB,16]`<br>;`c = indirect?mem[$rC,16]:$rC;`<br>`$rA = cmp_op(b,c);`         |
+| Operation   | `b = mem[$rB,16];`<br>`c = indirect?mem[$rC,16]:$rC;`<br>`$rA = cmp_op(b,c);`         |
 | Syntax      | `wdcm $rA, $rB, $rC, imm`                                                             |
 | Encoding    | `0x00 rA rB rC i`                                                                     |
 | Notes       |                                                                                       |
@@ -730,7 +730,7 @@ Panic if:
 |             |                                                                                       |
 |-------------|---------------------------------------------------------------------------------------|
 | Description | Compare or examine two 256-bit integers using selected mode                           |
-| Operation   | `b = mem[$rB,32]`<br>;`c = indirect?mem[$rC,32]:$rC;`<br>`$rA = cmp_op(b,c);`         |
+| Operation   | `b = mem[$rB,32];`<br>`c = indirect?mem[$rC,32]:$rC;`<br>`$rA = cmp_op(b,c);`         |
 | Syntax      | `wqcm $rA, $rB, $rC, imm`                                                             |
 | Encoding    | `0x00 rA rB rC i`                                                                     |
 | Notes       |                                                                                       |
--- a/src/vm/instruction_set.md
+++ b/src/vm/instruction_set.md
@@ -705,7 +705,7 @@ Then the actual operation that's performed:
 1     | ne   | Inequality (`!=`)
 2     | lt   | Less than (`<`)
 3     | gt   | Greater than (`>`)
-4     | lte  | Less than or equals (`>=`)
+4     | lte  | Less than or equals (`<=`)
 5     | gte  | Greater than or equals (`>=`)
 6     | lzc  | Leading zero count the lhs argument (`lzcnt`). Discards rhs.
 7     | -    | Reserved and must not be used
--- a/src/vm/instruction_set.md
+++ b/src/vm/instruction_set.md
@@ -688,8 +688,8 @@ Then the actual comparison that's performed:
 3     | gt   | Greater than (`>`)
 4     | le   | Less than or equals (`>=`)
 5     | ge   | Greater than or equals (`>=`)
-6     | -    | *Reserved*
-7     | -    | *Reserved*
+6     | -    | Reserved and must not be used
+7     | -    | Reserved and must not be used
 
 Clears `$of` and `$err`.
 
--- a/src/vm/instruction_set.md
+++ b/src/vm/instruction_set.md
@@ -663,7 +663,7 @@ Panic if:
 ### WDCM: 128-bit integer comparison
 
 |             |                                                                                       |
-|-------------|---------------------------------------------------------------------------------------| 
+|-------------|---------------------------------------------------------------------------------------|
 | Description | Compare two 128-bit integers using selected compare mode                              |
 | Operation   | `b = mem[$rB,16]`<br>;`c = indirect?mem[$rC,16]:$rC;`<br>`$rA = cmp_op(b,c);`         |
 | Syntax      | `wdcm $rA, $rB, $rC, imm`                                                             |
@@ -702,7 +702,7 @@ Panic if:
 ### WQCM: 256-bit integer comparison
 
 |             |                                                                                       |
-|-------------|---------------------------------------------------------------------------------------| 
+|-------------|---------------------------------------------------------------------------------------|
 | Description | Compare two 256-bit integers using selected compare mode                              |
 | Operation   | `b = mem[$rB,32]`<br>;`c = indirect?mem[$rC,32]:$rC;`<br>`$rA = cmp_op(b,c);`         |
 | Syntax      | `wqcm $rA, $rB, $rC, imm`                                                             |
@@ -719,11 +719,10 @@ Panic if:
 - `$rB + 16` overflows or `> VM_MAX_RAM`
 - `indirect == 1` and `$rC + 16` overflows or `> VM_MAX_RAM`
 
-
 ### WDOP: Misc 128-bit integer operations
 
 |             |                                                                                       |
-|-------------|---------------------------------------------------------------------------------------| 
+|-------------|---------------------------------------------------------------------------------------|
 | Description | Perform an ALU operation on two 128-bit integers                                      |
 | Operation   | `b = mem[$rB,16];`<br>`c = indirect?mem[$rC,16]:$rC;`<br>`mem[$rA,16] = op(b,c);`     |
 | Syntax      | `wdop $rA, $rB, $rC, imm`                                                             |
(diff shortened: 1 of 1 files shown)


src/vm/instruction_set.md
```diff
@@ -647,6 +660,299 @@ Panic if:
...
+| Operation   | `b = mem[$rB,16]`<br>;`c = indirect?mem[$rC,16]:$rC;`<br>`$rA = cmp_op(b,c);`         |
+| Syntax      | `wdcm $rA, $rB, $rC, imm`                                                             |
+| Encoding    | `0x00 rA rB rC i`                                                                     |
+| Notes       |                                                                                       |
+
+The six-bit immediate value is used to select operating mode, as follows:
+
+Bits     | Short name | Description
+---------|------------|-------------
+`...XXX` | `mode`     | Compre mode selection
```
mitch-fuel, line 677:
```suggestion
`...XXX` | `mode`     | Compare mode selection
```

src/vm/instruction_set.md
```diff
@@ -647,6 +660,299 @@ Panic if:
...
+
+The immediate value is interpreted identically to `WDCM`.
+
+Clears `$of` and `$err`.
+
+Panic if:
+
+- `$rA` is a [reserved register](./index.md#semantics)
+- `$rB + 16` overflows or `> VM_MAX_RAM`
+- `indirect == 1` and `$rC + 16` overflows or `> VM_MAX_RAM`
```
mitch-fuel, line 720:
```suggestion
- `$rB + 32` overflows or `> VM_MAX_RAM`
- `indirect == 1` and `$rC + 32` overflows or `> VM_MAX_RAM`
```

src/vm/instruction_set.md
```diff
@@ -647,6 +660,299 @@ Panic if:
...
+
+`$err` is cleared.
+
+Panic if:
+
+- The memory range `MEM[$rA, 16]`  does not pass [ownership check](./index.md#ownership)
+- `indirect0 == 1` and `$rB + 16` overflows or `> VM_MAX_RAM`
+- `indirect1 == 1` and `$rC + 16` overflows or `> VM_MAX_RAM`
+
+### WQML: Multiply 128-bit integers
```
mitch-fuel, line 809:
```suggestion
### WQML: Multiply 256-bit integers
```

src/vm/instruction_set.md
```diff
@@ -647,6 +660,299 @@ Panic if:
...
+
+|             |                                                                                       |
+|-------------|---------------------------------------------------------------------------------------|
+| Description | Perform integer multiplication operation on two 256-bit integers.                     |
+| Operation   | `b=indirect0?mem[$rB,16]:$rB;`<br>`c=indirect1?mem[$rC,16]:$rC;`<br>`mem[$rA,16]=b*c;`|
```
mitch-fuel, line 814:
```suggestion
### WQML: Multiply 256-bit integers

|             |                                                                                       |
|-------------|---------------------------------------------------------------------------------------|
| Description | Perform integer multiplication operation on two 256-bit integers.                     |
| Operation   | `b=indirect0?mem[$rB,32]:$rB;`<br>`c=indirect1?mem[$rC,32]:$rC;`<br>`mem[$rA,32]=b*c;`|
```

nfurfaro:
This looks good, and seems reasnonable. The only thing that would be nice to have in the initial pass at this would be a combined/fused muldiv op as that seems to be the most common case that requires users to bump up to a higher bit size integer.

>We could support these operations for 64-bit integers as well

This would be slick. `muldiv` for `u64` would be really useful.

pixelcircuits:
Looks good to me!

Dentosal:
I've added fused multiply-divide for 64bit registers as well, opcode `MLDV`. I'm still investigating whether exponentation and logarithm make sense to add.

src/vm/instruction_set.md
```diff
@@ -481,6 +497,22 @@ Panic if:
 
 `$err` is cleared.
 
+### MLDV: Fused multiply-divide
+
+|             |                                                                                       |
+|-------------|---------------------------------------------------------------------------------------|
+| Description | Multiplies two registers with arbitrary precision, then divides by a third register.  |
+| Operation   | `a = (b * c) / d;`                                                                    |
+| Syntax      | `wqdv $rA, $rB, $rC, $rD`                                                             |
```
Voxelot, line 506: This isn't a quadword op

```suggestion
| Syntax      | `muldiv $rA, $rB, $rC, $rD`                                                             |
```

Dentosal: Fixed

Dentosal:
@rostyslavtyshko: I've added leading-zeros operation to wdcm and wqcm, that can be used to implement log2 when combined with a subtract.

src/vm/instruction_set.md
```diff
@@ -647,6 +679,363 @@ Panic if:
...
+0    | add  | Add
+1    | sub  | Subtract
+2    | not  | Invert bits (discards rhs)
+3    | or   | Bitwise or
+4    | xor  | Bitwise exclusive or
+5    | and  | Bitwise and
+6    | shl  | Shift left (logical)
+7    | shr  | Shift right (logical)
+
+Operations behave `$of` and `$err` similarly to their 64-bit counterparts.
```
Voxelot, line 776: Should we call out that `$of` in this case would simply be set to true, instead of the overflow result which would be the case for 64-bit addition?

Dentosal: Well for addition the overflow is always either zero or one, since we can overflow one bit at most. However, bit shift right is affected by this. I'll update the comment.

src/vm/instruction_set.md
```diff
@@ -481,6 +497,22 @@ Panic if:
...
+
+|             |                                                                                       |
+|-------------|---------------------------------------------------------------------------------------|
+| Description | Multiplies two registers with arbitrary precision, then divides by a third register.  |
+| Operation   | `a = (b * c) / d;`                                                                    |
+| Syntax      | `mldv $rA, $rB, $rC, $rD`                                                             |
+| Encoding    | `0x00 rA rB rC rD`                                                                    |
+| Notes       | Division by zero is treated as division by `1 << 64` instead.                         |
+
+If the divisor (`$rD`) is zero, then instead the value is divided by `1 << 64`. This returns the higher half of the 128-bit multiplication result. This operation never overflows.
```
freesig, line 510: Won't this just return `0` if `(b * c) < (1 << 64)` or `1` if `(b * c) > (1 << 64)`

src/vm/instruction_set.md
```diff
@@ -647,6 +679,363 @@ Panic if:
 
 `$err` is cleared.
 
+### WDCM: 128-bit integer comparison
+
+|             |                                                                                       |
+|-------------|---------------------------------------------------------------------------------------|
+| Description | Compare or examine two 128-bit integers using selected mode                           |
+| Operation   | `b = mem[$rB,16]`<br>;`c = indirect?mem[$rC,16]:$rC;`<br>`$rA = cmp_op(b,c);`         |
```
freesig, line 687: I don't think this is rendering correctly

src/vm/instruction_set.md
```diff
@@ -647,6 +679,363 @@ Panic if:
...
+
+Then the actual operation that's performed:
+
+`mode`| Name | Description
+------|------|------------
+0     | eq   | Equality (`==`)
+1     | ne   | Inequality (`!=`)
+2     | lt   | Less than (`<`)
+3     | gt   | Greater than (`>`)
+4     | lte  | Less than or equals (`>=`)
```
freesig, line 708:
```suggestion
4     | lte  | Less than or equals (`<=`)
```

src/vm/instruction_set.md
```diff
@@ -481,6 +497,22 @@ Panic if:
 
 `$err` is cleared.
 
+### MLDV: Fused multiply-divide
+
+|             |                                                                                       |
+|-------------|---------------------------------------------------------------------------------------|
+| Description | Multiplies two registers with arbitrary precision, then divides by a third register.  |
```
freesig, line 504: What does `arbitrary precision` refer to here? Does it mean the max value would be `u64::MAX * u64::MAX`?

freesig (commented):
I wonder if we want to also returning the remainders as well as the Quotient for some of the division?

Dentosal:
> I wonder if we want to also returning the remainders as well as the Quotient for some of the division?

@freesig Remainder can be obtained using `addmod(dividend, 0, divisor)`.

src/vm/instruction_set.md
Dentosal: Fixed.

src/vm/instruction_set.md
Dentosal: No. For instance `(u64::MAX*u64::MAX) / (1 << 64) =18446744073709551614 `

nfurfaro (approved):

PR #486, opened 2023-04-29
Bitshift should discard overflow

Dentosal:
Bit shift to right was documented to set `$of` in case of underflow. This behavior is both unintuitive and different from the current implementation. See also https://github.com/FuelLabs/fuel-vm/pull/433 for a related bugfix.


mitch-fuel (approved):
this seems clear to me

freesig (approved):

PR #491, opened 2023-06-05
Granular Asset Minting

SilentCicero:
### Abstract

This PR introduces granular asset minting which would enable sub identifiers to be included at asset creation. These sub identifiers can include important metadata and be used to enable native level NFTs.

**This adds:**
- A new asset ID construction (which incorporates the contract ID and the sub asset ID).
- New receipt logs for minting and burning which are helpful for indexing.
- Changes to the MINT and BURN opcodes.

### Cons

- This removes the nice feature that asset IDs are the contract ID.

### Pros

- Native level UTXO based NFTs.
- Single contract multi-asset pools (useful for common AMM designs).

Implementation tickets/PRs:

- https://github.com/FuelLabs/fuel-vm/issues/496
- https://github.com/FuelLabs/fuels-ts/issues/1067
- https://github.com/FuelLabs/fuel-indexer/issues/1012
- https://github.com/FuelLabs/sway/pull/4651


src/vm/instruction_set.md
```diff
@@ -1795,28 +1808,43 @@ Logs the memory range `MEM[$rC, $rD]`.
...
+| Syntax      | `mint $rA $rB`                                       |
+| Encoding    | `0x00 rA rB - -`                                     |
+| Notes       | `$rB` is a pointer to a 256-bit ID in memory         |
+
+The asset ID will be constructed using the coin ID construction method.
 
 Panic if:
 
+- If asset ID has already been instantiated by another contract.
+- `$rD > VM_MAX_RAM`
```
Dentosal, line 1829: I think this should be `$rB + 32 > VM_MAX_RAM`

Dentosal:
Possible optimization would be to add a cheap way to get the default subasset, i.e. all zeroes. It would be quite unfortunate if ALOC+CLI instruction pair must be performed even when a contract doesn't use subassets. A possible downside with this approach is that it's possible to not support subassets in some contract, but I think that can be alleviated by having proper tools on Sway side.

I think we can merge this as-is anyway, and figure out what, if any, optimizations are needed when implementing this.

dmihal:
I was just looking over [EIP-3525](https://eips.ethereum.org/EIPS/eip-3525), which uses the word "slot" to represent the sub-IDs of these tokens.

Any thoughts on the word "slot" instead of "sub_id"?

@SilentCicero @SwayStar123

SwayStar123:
> 



> I was just looking over [EIP-3525](https://eips.ethereum.org/EIPS/eip-3525), which uses the word "slot" to represent the sub-IDs of these tokens.
> 
> Any thoughts on the word "slot" instead of "sub_id"?
> 
> @SilentCicero @SwayStar123

Slots and token_ids seem to be seperate concepts in erc3525, I dont think it would be accurate here. The original multi-token rfc that I wrote was inspired by erc1155.

I would prefer just identifier over sub_indentifier, as contractid and assetid are already distinct i dont think there would be any confusion here, but sub_indentifier is also fine

src/protocol/id/asset.md
```diff
@@ -0,0 +1,8 @@
+# Asset ID
+
+The _asset ID_ (also called _asset hash_) of a asset is computed as
+the [hash](../cryptographic_primitives.md#hashing) of the `CONTRACT_ID` and a 256-bit `SUB_IDENTIFIER`.
```
xgreenx, line 4: Do we want to support the `None` sub-identifier? In this case, the contract still can mint `AssetId == ContractId`.

Of course, we can't use `None`, but we can use some magic value instead like zeroed sub-identifier or a maximum sub-identifier.

src/protocol/abi/receipts.md
```diff
@@ -23,6 +23,8 @@ All receipts will have a `type` property:
   - [Panic](#panic-receipt)
   - [Revert](#revert-receipt)
   - [Log](#log-receipt)
+  - [Mint](#mint-receipt)
```
xgreenx, line 26: We also need to update `ReceiptType` in the `src/protocol/tx_format/transaction.md`

SilentCicero:
> Possible optimization would be to add a cheap way to get the default subasset, i.e. all zeroes. It would be quite unfortunate if ALOC+CLI instruction pair must be performed even when a contract doesn't use subassets. A possible downside with this approach is that it's possible to not support subassets in some contract, but I think that can be alleviated by having proper tools on Sway side.
> 
> I think we can merge this as-is anyway, and figure out what, if any, optimizations are needed when implementing this.

Regarding this, there might be some clever ways we can tuck the pre-image in the transaction.

src/vm/instruction_set.md
```diff
@@ -1567,11 +1567,13 @@ Block header hashes for blocks with height greater than or equal to current bloc
...
-| Notes       |                                                      |
+| Description | Burn `$rA` coins of the `$rB` ID from the current contract. |
+| Operation   | ```burn($rA, $rB);```                                |
+| Syntax      | `burn $rA $rB`                                       |
+| Encoding    | `0x00 rA rB - -`                                     |
+| Notes       | `$rB` is a pointer to a 256-bit ID in memory.        |
+
+The asset ID is constructed using the coin ID construction method.
 
 Panic if:
```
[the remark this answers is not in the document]
Voxelot: updated to use a helper function to represent the asset id construction function

src/protocol/id/asset.md
dmihal: I believe the decision was to rip the band-aid off, and break the tie that `AssetId == ContractID`

It'll be more confusing if they're _sometimes_ equal, and sometimes different.

xgreenx: If we don't need backward compatibility, then okay=) But it means it is a super breaking change

dmihal: Yep, it's a breaking change, but that's why it's important we get it in soon so teams can adjust

xgreenx (dismissed):

src/abi/receipts.md
```diff
@@ -124,6 +126,46 @@ _Important note:_ For the JSON representation of receipts, we represent 64-bit u
...
+- `contract_id`: Hexadecimal string representation of the 256-bit (32-byte) contract ID of the current context.
+- `val`: Decimal string representation of a 64-bit unsigned integer; value of register `$rA`.
+- `pc`: Hexadecimal string representation of a 64-bit unsigned integer; value of register `$pc`.
+- `is`: Hexadecimal string representation of a 64-bit unsigned integer; value of register `$is`.
+
+```json
+{
+  "type": "Mint",
+  "sub_id": "0x39150017c9e38e5e280432d546fae345d6ce6d8fe4710162c2e3a95a6faff051",
+  "contract_id": "0x39150017c9e38e5e280432d546fae345d6ce6d8fe4710162c2e3a95a6faff051",
```
Voxelot, line 142: this breaks our convention of using `id` for the contract like other receipts. Any particular reason we used contract_id instead of just `id`? cc @SilentCicero @dmihal

Voxelot (approved):

dmihal:
Any thoughts on having `MINT` and `BURN` return the resulting asset ID? Would save some contracts from needing to do duplicate hashing to calculate the address.

PR #495, opened 2023-06-09
docs: SMT Delete interface takes leaf data

bvrooman:
Related issues:
 - https://github.com/FuelLabs/fuel-vm/issues/472

We are changing the SMT `delete` signature to take in both the original leaf key and leaf data. This is to address a potential collision when inserting leaves. See the discussion in the above related issues for details.


bvrooman:
Fixed the issue without API change!

PR #498, opened 2023-06-20
New OutputData type

SilentCicero:
## Abstract

This output is for including arbitrary data within a transaction. This allows users and developers to include arbitrary data that can be signed over that cannot be included in a script, input, smart contract or within predicate code.

Note, this output is not spendable.

## Prior Art

https://en.bitcoin.it/wiki/OP_RETURN

### DIFF
--- a/src/protocol/tx_format/output.md
+++ b/src/protocol/tx_format/output.md
@@ -7,6 +7,7 @@ enum OutputType : uint8 {
     Change = 2,
     Variable = 3,
     ContractCreated = 4,
+    Data = 5,
 }
 ```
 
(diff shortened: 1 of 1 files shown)


xgreenx:
Why can't we use `predicate_data` or `script_data` to store arbitrary data?

> One potential use case for an output type like this would be providing the preimage to an asset ID to determine the contract ID without having to annoyingly include that data as an argument in a deposit method for things like AMMs. Instead, you can have a function included in the standard library which just scans the outputs for this preimage in a specific output data.

People can still use this new output to store different kinds of data(not only the preimage of the asset id), so SDK should understand the purpose of each output somehow. I think that if your code requires a preimage of the asset id, then you should provide it as an argument. Otherwise, it is chaos with random outputs.

PR #499, opened 2023-06-21
Add scattered push/pop operations

### LINKED ISSUE #407: Add PUSH and POP instructions.
### Use case

When creating a local function call frame (as opposed to a more global contract call frame) the compiler must preserve the value of usually several general purpose registers to avoid clobbering values in the caller.

It is currently (not merged yet, but soon) doing this manually by copying `$SP`, using `CFEI` and performing several `STORE`s, potentially up to ~40 in the worst case at the start of a function, and reversing the process with up to ~40 `LOAD`s at the end.

Even fairly small functions may use around 8 GP regs which would equate to 20 instructions to save and restore them, or 80 bytes of binary executable.  Per function.

### Proposal

I propose we introduce `PUSH` and `POP` to do all this with single instructions, saving a lot of wasted bytecode.  80 bytes becomes 8 bytes.  A difference this significant could determine whether a function is inlined or not and so has further implications for optimisation.

The arguments could be a range of registers, first to last inclusive.  `PUSH $r0 $r3` would save all 4 of `$r0`, `$r1`, `$r2` and `$r3` to the stack.  `POP $r0 $r3` would restore them correctly.

I assume it's possible for the gas cost to just remain the equivalent of a `CFEI` followed by _n_ `STORE` instructions.

Dentosal:
Initial PR to gather some feedback on push/pop instructions. Closes  #407.

There's a possible further improvement at the cost of simplicity: we could special-case pushing zero values (immediate bitmask all zeroes) to do something useful instead. For instance, it could be used to push/pop all user registers using a single operation. However I feel like that wouldn't be too useful.


vaivaswatha:
Is the cost of this instruction going to depend on how many registers were pushed? 

LGTM.

IGI-111:
LGTM as well

Dentosal:
> Is the cost of this instruction going to depend on how many registers were pushed?

No. It's cheap enough to execute anyway, the cost between 1 and 24 pushed registers is likely smaller than per-instruction overhead.

vaivaswatha (approved):

Voxelot:
One thing I'm wondering about is how safe it is to assume that the stack will be properly cleared after an internal function call before popping the registers back. Otherwise there's a risk that we'd be popping arbitrary data into the registers if the stack wasn't restored to its original state while pushing the registers. While this is a sway implementation concern, let's say I write a sway function that has an asm block which performs a `cfe`, would sway be able to properly restore the stack to it's original state before the registers are popped while returning from that fn? @vaivaswatha

vaivaswatha:
> One thing I'm wondering about is how safe it is to assume that the stack will be properly cleared after an internal function call before popping the registers back. Otherwise there's a risk that we'd be popping arbitrary data into the registers if the stack wasn't restored to its original state while pushing the registers. While this is a sway implementation concern, let's say I write a sway function that has an asm block which manually mucks with the stack by performing a `cfe`, would sway be able to properly restore the stack to its original state before the registers are popped while returning from that fn? @vaivaswatha

Isn't this a concern today as well? where we push the registers to the stack before a call and pop them afterwards, but just with separate instructions instead of this one single one.

I'd say it's hard for the compiler to do anything safely if ASM blocks are used unsafely.

Voxelot:
One workaround I can think of is if the compiler also pushed $sp to the stack before a function call, and after returning it shrinks the stack back to that value automatically before any popping.

PR #508, opened 2023-08-02
Remove unused MEM_MAX_ACCESS_SIZE

Dentosal:
The configuration value for it was fixed to `MEM_MAX_ACCESS_SIZE = VM_MAX_RAM`. The rule against exceeding `VM_MAX_RAM` always triggers first. See https://github.com/FuelLabs/fuel-vm/pull/511#discussion_r1276819690 and the PR in general for the removal from `fuel-vm`.


Voxelot:
I think it's fine to leave this out for now since we haven't found a practical use for it yet. We can always add it back in later if it's deemed necessary for the prover.

Voxelot (approved):

PR #511, opened 2023-08-15
Change mint tx format and coinbase rules to mitigate dust accumulation

Voxelot:
The mint transaction is no longer allowed to create coin outputs, and instead performs an intrinsic contract balance increase.

This is lighterweight compared to using a script transaction as a coinbase, as no VM execution needs to occur. This prevents dust accumulation of low-value coinbase coin outputs compared to the previous design.

### DIFF
--- a/src/protocol/tx-validity.md
+++ b/src/protocol/tx-validity.md
@@ -107,10 +107,9 @@ def sum_inputs(tx, asset_id) -> int:
 Returns any minted amounts by the transaction
 """
 def minted(tx, asset_id) -> int:
-    mint_amount: int = 0
-    if tx.type == TransactionType.Mint and asset_id == tx.mint_asset_id:
-        mint_amount = tx.mint_amount
-    return mint_amount
+    if tx.type != TransactionType.Mint or asset_id != tx.mint_asset_id:
+        return 0
+    return tx.mint_amount
 
 def sum_outputs(tx, asset_id) -> int:
     total: int = 0
--- a/src/protocol/tx-validity.md
+++ b/src/protocol/tx-validity.md
@@ -107,10 +107,10 @@ def sum_inputs(tx, asset_id) -> int:
 Returns any minted amounts by the transaction
 """
 def minted(tx, asset_id) -> int:
-  mint_amount: int = 0
-  if tx.type == TransactionType.Mint and asset_id == tx.mint_asset_id:
-    mint_amount = tx.mint_amount
-  return mint_amount
+    mint_amount: int = 0
+    if tx.type == TransactionType.Mint and asset_id == tx.mint_asset_id:
+        mint_amount = tx.mint_amount
+    return mint_amount
 
 def sum_outputs(tx, asset_id) -> int:
     total: int = 0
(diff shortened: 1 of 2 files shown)


src/protocol/tx-validity.md
```diff
@@ -206,11 +215,25 @@ Transaction processing is completed by removing spent UTXOs from the state and a
...
-The coinbase transaction is a mechanism for block creators to convert fees into spendable UTXOs.
+The coinbase transaction is a mechanism for block creators to collect transaction fees.
 
 In order for a coinbase transaction to be valid:
 
 1. It must be a [Mint](../tx-format/transaction.md#TransactionMint) transaction.
-2. The coinbase transaction must be the first transaction within a block, even if there are no other transactions in the block and the fee is zero.
-3. The total output value of the coinbase transaction cannot exceed the total amount of fees processed from all other transactions within the same block.
-4. The `asset_id` for coinbase transaction outputs must match the `asset_id` that fees are paid in (`asset_id == 0`).
+1. The coinbase transaction must be the first transaction within a block, even if there are no other transactions in the block and the fee is zero.
```
xgreenx, line 223: It should be the last transaction in the block to avoid affecting other transactions.

src/tx-format/transaction.md
```diff
@@ -131,15 +131,25 @@ Creates a contract with contract ID as computed [here](../identifiers/contract-i
...
-|----------------|------------------------------|------------------------------------------------------|
-| `txPointer`    | [TXPointer](./tx-pointer.md) | The location of the `Mint` transaction in the block. |
-| `outputsCount` | `uint8`                      | Number of outputs.                                   |
-| `outputs`      | [Output](./output.md)`[]`    | List of outputs.                                     |
+This means that the transaction ID must also include the correct `txPointer` value, not zeroed out.
+
+| name           | type                         | description                                               |
+|----------------|------------------------------|-----------------------------------------------------------|
+| `mintAmount`   | `uint64`                     | The amount of funds being minted within this transaction. |
+| `mintAssetId`  | `byte[32]`                   | The asset ID of funds minted within this transaction.     |
```
xgreenx, line 139: It may make sense to have an array of the mint assets. In the case if we want to support paying in different assets=)

src/tx-format/transaction.md
```diff
@@ -131,15 +131,25 @@ Creates a contract with contract ID as computed [here](../identifiers/contract-i
...
+| `gasLimit`     | `uint64`                     | Gas limit for transaction                                 |
```
xgreenx, line 140: If we don't have gas price, it means mint transaction is free. So maybe it is better if the gas limit is part of the `ConsensusParams` instead of a customizable value?

src/tx-format/transaction.md
```diff
@@ -131,15 +131,25 @@ Creates a contract with contract ID as computed [here](../identifiers/contract-i
...
+| `scriptLength` | `uint16`                     | Script length, in instructions.                           |
+| `inputsCount`  | `uint8`                      | Number of inputs.                                         |
+| `outputsCount` | `uint8`                      | Number of outputs.                                        |
+| `txPointer`    | [TXPointer](./tx-pointer.md) | The location of the `Mint` transaction in the block.      |
+| `receiptsRoot` | `byte[32]`                   | Merkle root of receipts.                                  |
+| `script`       | `byte[]`                     | Script to execute.                                        |
```
xgreenx, line 146: If we allow running any script, the `Mint` transaction may fail. Do we need to worry about that, or is it okay?=)

src/tx-format/transaction.md
Voxelot: Yeah that makes sense, we'd essentially be saying there's a max limit of execution time a producer is allowed to spend on their coinbase processing agreed upon by all chain participants.

src/tx-format/transaction.md
Voxelot: Hmm good point - idk if there's really a way around this, especially if they use too much gas. Worst case they can put a change output to capture the fees if the script reverts, even if it's dust.

src/protocol/tx-validity.md
Voxelot: Fine with me

src/tx-format/transaction.md
Voxelot: I thought about that, and it kind of made me wonder if it'd be better to just replace the concept of a mint transaction with a mint input that has special spending rules.

Eventually that could be opened up later to enable things like flashloans etc.

src/protocol/tx-validity.md
Voxelot: updated

src/tx-format/transaction.md
Voxelot: removed all scripting from mint tx

src/tx-format/transaction.md
Voxelot: decided to keep it simple with a single assetID

src/tx-format/transaction.md
Voxelot: removed the script and switched to intrinsic balance increase

src/protocol/tx-validity.md
```diff
@@ -103,6 +103,15 @@ def sum_inputs(tx, asset_id) -> int:
...
     return total
 
+"""
+Returns any minted amounts by the transaction
+"""
+def minted(tx, asset_id) -> int:
+    mint_amount: int = 0
+    if tx.type == TransactionType.Mint and asset_id == tx.mint_asset_id:
+        mint_amount = tx.mint_amount
+    return mint_amount
```
Dentosal, line 113:
```suggestion
def minted(tx, asset_id) -> int:
    if tx.type != TransactionType.Mint or asset_id != tx.mint_asset_id:
        return 0
    return tx.mint_amount
```

src/tx-format/transaction.md
```diff
@@ -131,15 +131,18 @@ Creates a contract with contract ID as computed [here](../identifiers/contract-i
...
-| `txPointer`    | [TXPointer](./tx-pointer.md) | The location of the `Mint` transaction in the block. |
-| `outputsCount` | `uint8`                      | Number of outputs.                                   |
-| `outputs`      | [Output](./output.md)`[]`    | List of outputs.                                     |
+| name             | type                         | description                                          |
+|------------------|------------------------------|------------------------------------------------------|
+| `txPointer`      | [TXPointer](./tx-pointer.md) | The location of the `Mint` transaction in the block. |
+| `inputContract`  | [Input](./input.md)          | The contract utxo that assets are minted to.         |
+| `outputContract` | [Output](./output.md)        | The contract utxo that assets are being minted to.   |
+| `mintAmount`     | `uint64`                     | The amount of funds minted.                          |
+| `mintAssetId`    | `byte[32]`                   | The asset IDs corresponding to the minted amount.    |
```
xgreenx, line 142: Maybe it is better to keep the receipt root and produce burn and mint receipts for the indexer(as you mentioned in Slack). Otherwise, it requires additional handling from other parts of our stack.

xgreenx: Let's keep it simple for now

xgreenx (approved):

xgreenx:
We also need to develop a Sway contract that can be used by the block producer

PR #514, opened 2023-08-24
Transaction Policies

Voxelot:
# Summary

Use a dynamic set of policies for transaction validation requirements such as gaslimit & maturity. This will reduce the size of transactions that always leave these fields unset, and also future proof the transaction format for later requirements such as multi-dimensional pricing.

A new policy that's been added is `witnessLimit`. Having a witness limit allows the network to charge for witness data without exposing users to malicious basefee manipulation by third-parties / relayers. Without a charge & limit for witness data, blockspace could be easily saturated with spurious spammy witness data.

Another significant change is to separate base costs of the transaction from the transaction gas limit. Gas limit will now only apply to script execution, and everything else is charged as part of the minimum transaction fee. 

Since the gas limit can't restrict the total cost of the transaction anymore, another policy called `MaxFee` was added. This is  a more direct way to place an upper bound on transaction costs and is also future compatible with multi-dimensional pricing.

## Multi-dimension pricing preparation

In order for resources such as execution and storage to be charged for independently, transactions must be able to specify an anticipated consumption limit for each resource kind. By abstracting these limits into a dynamic set of policies, new resource types can be added in the future more easily.


src/tx-format/policy.md
```diff
@@ -0,0 +1,51 @@
...
+
+Transaction is invalid if:
+
+- `type > PolicyType.Maturity`
+
+## GasLimit
+
+| name       | type     | description                                         |
+|------------|----------|-----------------------------------------------------|
+| `gasLimit` | `uint64` | Gas limit for transaction (including predicate gas) |
```
xgreenx, line 24: I think we can exclude predicate gas from the gas limit and make it entirely dedicated to the script execution.

We can treat the predicate gas as part of the minimum fee because it is the exact value that we plan to charge from the user, while the script execution depends on the state of the blockchain and may be different -> affects the maximum fee.

src/tx-format/transaction.md
```diff
@@ -51,27 +49,27 @@ enum ReceiptType : uint8 {
...
-| `scriptData`       | `byte[]`                    | Script input data (parameters).                      |
-| `inputs`           | [Input](./input.md)`[]`     | List of inputs.                                      |
-| `outputs`          | [Output](./output.md)`[]`   | List of outputs.                                     |
-| `witnesses`        | [Witness](./witness.md)`[]` | List of witnesses.                                   |
+| name               | type                        | description                            |
+|--------------------|-----------------------------|----------------------------------------|
+| `gasPrice`         | `uint64`                    | Gas price for transaction.             |
+| `scriptLength`     | `uint16`                    | Script length, in instructions.        |
+| `scriptDataLength` | `uint16`                    | Length of script input data, in bytes. |
+| `policyCount`      | `uint8`                     | Number of policies.                    |
```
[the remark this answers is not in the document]
Voxelot: Ah good point, I thought I had them ordered correctly but I guess not

src/tx-format/transaction.md
```diff
@@ -51,27 +49,27 @@ enum ReceiptType : uint8 {
...
-| `witnessesCount`   | `uint8`                     | Number of witnesses.                                 |
-| `receiptsRoot`     | `byte[32]`                  | Merkle root of receipts.                             |
-| `script`           | `byte[]`                    | Script to execute.                                   |
-| `scriptData`       | `byte[]`                    | Script input data (parameters).                      |
-| `inputs`           | [Input](./input.md)`[]`     | List of inputs.                                      |
-| `outputs`          | [Output](./output.md)`[]`   | List of outputs.                                     |
-| `witnesses`        | [Witness](./witness.md)`[]` | List of witnesses.                                   |
+| name               | type                        | description                            |
+|--------------------|-----------------------------|----------------------------------------|
+| `gasPrice`         | `uint64`                    | Gas price for transaction.             |
```
xgreenx, line 59: Do we also want to move `gas_price` to policies as part of the preparation for the multi-dimension prices?

With the current approach, we will include it in each transaction, but later, with storage price and on-chain gas price, we can remove it or treat it like a tip.

xgreenx:
While we move some of the fields to policies, we don't plan to change the API of the `fuel-tx` and `GraphQL` to minimize the breaks of the change. Getters and setters remain the same.

The changes will affect the serialization and deserialization of the transactions, and a minimal fee to include `WhitnessLimit`.

What should be changed to support policies:
- `fuel-tx`:
    - Add a new `WhitnessLimit` getter that returns `None<u64>`.
    - Add a new enum for policies.
    - Add iteration over policies.
    - Old API keeps the same. Only the implementation looks into `policies`.
    - The minimal fee considers a `WhitnessLimit` if it is set.

- `fuel-core`:
    - Add support for a new enum in the schema.
    - Add support for policies on the GraphQL level.
    - Old API keeps the same. Only the implementation looks into `policies`.
    - Maybe reject transactions on the TxPool level if `WhitnessLimit` is not set.
    - Maybe reject transactions without `WhitnessLimit` during `dry_run`.

- `fuels-rs`:
    - Add support for `WhitnessLimit`. It includes automatically setting this field before signing if the user does not set it(maybe for simplicity, we could use some constant).
    - Support new enums from schema.
    - Maybe provide access to policies to the user.
    

- `fuels-ts`:
    - If https://github.com/FuelLabs/fuel-vm/issues/579 is not ready and TS SDK doesn't use implementation of the serialization/deserialization from the `fuel-tx`, then we need to update se/de for `Script` and `Create` transactions.
    - Add support for new enum in the schema.
    - Add support for `WhitnessLimit`, including automating its setting before signing if it is not set by the user.
    - Add manual access to policies.
    - Keep the old API the same, but work with policies instead of fields.

src/tx-format/transaction.md
Voxelot: true, moving it here would make it easier to deprecate later on with an automated pricing model.

Voxelot:
I don't think specifying a witness limit should be mandatory at the protocol level, but maybe the SDK's should validate this for users.

Voxelot:
I also realized this PR needs to update the GTF opcodes, although I don't think this will impact sway a lot except for some additive changes to the std-lib.

Voxelot:
@xgreenx I've updated the PR to move gasPrice to policies, added requirements to tx-validity specifying that the gasPrice policy is required for now, and also added policies to the GTF args.

src/tx-format/transaction.md
xgreenx: The field is still here=D

src/tx-format/policy.md
```diff
@@ -0,0 +1,57 @@
...
+
+| name           | type     | description                                                    |
+|----------------|----------|----------------------------------------------------------------|
+| `witnessLimit` | `uint64` | The maximum amount of witness data allowed for the transaction |
+
+Given helper `len()` that returns the number of bytes of a field.
+
+Transaction is invalid if:
+
+- `witnessLimit > len(tx.witnesses)`
```
xgreenx, line 47:
```suggestion
- `len(tx.witnesses) > witnessLimit`
```

src/fuel-vm/instruction-set.md
```diff
@@ -2425,77 +2425,77 @@ Get [fields from the transaction](../tx-format/transaction.md).
...
+| `GTF_OUTPUT_COIN_AMOUNT`                  | `0x302` | `tx.outputs[$rB].amount`                         |
+| `GTF_OUTPUT_COIN_ASSET_ID`                | `0x303` | Memory address of `tx.outputs[$rB].asset_id`     |
+| `GTF_OUTPUT_CONTRACT_INPUT_INDEX`         | `0x304` | `tx.outputs[$rB].inputIndex`                     |
+| `GTF_OUTPUT_CONTRACT_BALANCE_ROOT`        | `0x305` | Memory address of `tx.outputs[$rB].balanceRoot`  |
+| `GTF_OUTPUT_CONTRACT_STATE_ROOT`          | `0x306` | Memory address of `tx.outputs[$rB].stateRoot`    |
+| `GTF_OUTPUT_CONTRACT_CREATED_CONTRACT_ID` | `0x307` | Memory address of `tx.outputs[$rB].contractID`   |
+| `GTF_OUTPUT_CONTRACT_CREATED_STATE_ROOT`  | `0x308` | Memory address of `tx.outputs[$rB].stateRoot`    |
+| `GTF_WITNESS_DATA_LENGTH`                 | `0x400` | `tx.witnesses[$rB].dataLength`                   |
+| `GTF_WITNESS_DATA`                        | `0x401` | Memory address of `tx.witnesses[$rB].data`       |
+| `GTF_POLICY_GAS_PRICE`                    | `0x500` | `tx.policies[0x00].gasPrice`                     |
```
xgreenx, line 2495: Do we need to provide access to policies itself? Like count of policies and policy at index?

src/fuel-vm/instruction-set.md
```diff
@@ -2425,77 +2425,77 @@ Get [fields from the transaction](../tx-format/transaction.md).
...
-| `GTF_OUTPUT_COIN_AMOUNT`                  | `0x203` | `tx.outputs[$rB].amount`                         |
-| `GTF_OUTPUT_COIN_ASSET_ID`                | `0x204` | Memory address of `tx.outputs[$rB].asset_id`     |
-| `GTF_OUTPUT_CONTRACT_INPUT_INDEX`         | `0x205` | `tx.outputs[$rB].inputIndex`                     |
-| `GTF_OUTPUT_CONTRACT_BALANCE_ROOT`        | `0x206` | Memory address of `tx.outputs[$rB].balanceRoot`  |
-| `GTF_OUTPUT_CONTRACT_STATE_ROOT`          | `0x207` | Memory address of `tx.outputs[$rB].stateRoot`    |
-| `GTF_OUTPUT_CONTRACT_CREATED_CONTRACT_ID` | `0x208` | Memory address of `tx.outputs[$rB].contractID`   |
-| `GTF_OUTPUT_CONTRACT_CREATED_STATE_ROOT`  | `0x209` | Memory address of `tx.outputs[$rB].stateRoot`    |
-| `GTF_WITNESS_DATA_LENGTH`                 | `0x301` | `tx.witnesses[$rB].dataLength`                   |
-| `GTF_WITNESS_DATA`                        | `0x302` | Memory address of `tx.witnesses[$rB].data`       |
+| `GTF_SCRIPT_MATURITY`                     | `0x002` | `tx.maturity`                                    |
```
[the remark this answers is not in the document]
Voxelot: ah i should remove it

src/fuel-vm/instruction-set.md
Voxelot: good idea

Voxelot:
@xgreenx fixed the comments you left, and pushed a separate commit for a different idea I had about using bitfields to index policies and reduce the amount of data that policies consume.

src/tx-format/transaction.md
```diff
@@ -51,29 +49,31 @@ enum ReceiptType : uint8 {
...
-| `script`           | `byte[]`                    | Script to execute.                                   |
-| `scriptData`       | `byte[]`                    | Script input data (parameters).                      |
-| `inputs`           | [Input](./input.md)`[]`     | List of inputs.                                      |
-| `outputs`          | [Output](./output.md)`[]`   | List of outputs.                                     |
-| `witnesses`        | [Witness](./witness.md)`[]` | List of witnesses.                                   |
+| name               | type                        | description                            |
+|--------------------|-----------------------------|----------------------------------------|
+| `scriptLength`     | `uint16`                    | Script length, in instructions.        |
+| `scriptDataLength` | `uint16`                    | Length of script input data, in bytes. |
+| `policyTypes`      | `uint32`                    | Bitfield of used policy types.         |
```
xgreenx, line 61: I think it's okay to use `u64` here since it takes 8 bytes in canonical format.

```suggestion
| `policyTypes`      | `uint64`                    | Bitfield of used policy types.         |
```

src/tx-format/transaction.md
```diff
@@ -94,22 +98,25 @@ The receipts root `receiptsRoot` is the root of the [binary Merkle tree](../prot
...
-| `salt`                 | `byte[32]`                  | Salt.                                                |
-| `storageSlots`         | `(byte[32], byte[32]])[]`   | List of storage slots to initialize (key, value).    |
-| `inputs`               | [Input](./input.md)`[]`     | List of inputs.                                      |
-| `outputs`              | [Output](./output.md)`[]`   | List of outputs.                                     |
-| `witnesses`            | [Witness](./witness.md)`[]` | List of witnesses.                                   |
+| name                   | type                        | description                                       |
+|------------------------|-----------------------------|---------------------------------------------------|
+| `bytecodeLength`       | `uint16`                    | Contract bytecode length, in instructions.        |
+| `bytecodeWitnessIndex` | `uint8`                     | Witness index of contract bytecode to create.     |
+| `policyCount`          | `uint8`                     | Number of policies.                               |
```
xgreenx, line 105:
```suggestion
| `policyTypes`      | `uint64`                    | Bitfield of used policy types.         |
```

xgreenx (commented):
I like the idea of using bitmask as the length provider and type of policy provider. It saves 8 bytes per policy. It makes implementation a little bit complicated, but it is worth of that.

src/tx-format/transaction.md
Voxelot: True but it could save data outside of the VM for things like DA

src/tx-format/policy.md
```diff
@@ -0,0 +1,65 @@
...
+
+Transaction is invalid if:
+
+- `blockheight() < maturity`
+
+## MaxFee
+
+| name      | type     | description                                  |
+|-----------|----------|----------------------------------------------|
+| `max_fee` | `uint64` | The maximum fee payable by this transaction. |
```
xgreenx, line 60: Is that value in the base token, or is that value in gas? 

- Base token is good because we can later include storage price there too, but there can be a problem with alternative tokens that allow paying for gas(if we can have more than one token within the transaction to do that).

- In gas is good because it is agnostic to the token, but we will face the problem when we introduce a separate storage limit and storage price.

So it seems the first option with value in the base token and rule that we can't use different assets to pay for the transaction.

src/fuel-vm/instruction-set.md
```diff
@@ -2439,86 +2439,87 @@ Set `$rA` to the index of the currently-verifying predicate.
...
+| `GTF_OUTPUT_CONTRACT_CREATED_STATE_ROOT`  | `0x308` | Memory address of `tx.outputs[$rB].stateRoot`                      |
+| `GTF_WITNESS_DATA_LENGTH`                 | `0x400` | `tx.witnesses[$rB].dataLength`                                     |
+| `GTF_WITNESS_DATA`                        | `0x401` | Memory address of `tx.witnesses[$rB].data`                         |
+| `GTF_POLICY_COUNT`                        | `0x500` | `len(tx.policies)`                                                 |
+| `GTF_POLICY_TYPE`                         | `0x501` | `tx.policies[$rB].type`                                            |
+| `GTF_POLICY_GAS_PRICE`                    | `0x502` | `tx.policies[0x00].gasPrice`                                       |
+| `GTF_POLICY_GAS_LIMIT`                    | `0x503` | `tx.policies[count_ones(0x11 & tx.policyTypes) - 1].gasLimit`      |
+| `GTF_POLICY_WITNESS_LIMIT`                | `0x504` | `tx.policies[count_ones(0x111 & tx.policyTypes) - 1].witnessLimit` |
+| `GTF_POLICY_MATURITY`                     | `0x505` | `tx.policies[count_ones(0x1111 & tx.policyTypes) - 1].maturity`    |
```
xgreenx, line 2515:
```suggestion
| `GTF_POLICY_MATURITY`                     | `0x505` | `tx.policies[count_ones(0x1111 & tx.policyTypes) - 1].maturity`    |
| `GTF_POLICY_MAX_FEE`                     | `0x506` | `tx.policies[count_ones(0x11111 & tx.policyTypes) - 1].maxFee`    |
```

src/tx-format/transaction.md
```diff
@@ -83,7 +83,11 @@ Transaction is invalid if:
 - `scriptDataLength > MAX_SCRIPT_DATA_LENGTH`
 - `scriptLength * 4 != len(script)`
 - `scriptDataLength != len(scriptData)`
-- `gasLimit` is less than the sum of all `predicateGasUsed` for `InputType.Coin` or `InputType.Message` where predicate length is greater than zero.
+- No policy of type `PolicyType.GasPrice`
+- No policy of type `PolicyType.GasLimit`
```
xgreenx, line 87: I think we can allow not have`PolicyType.GasLimit` in the case of an empty `script`.

xgreenx (commented):
LGTM!

luizstacio:
# Frontend:

The `fuels-ts` support will block us, which means our timeline will be additional to the time they need to implement first.

 - Fuel Wallet
    - [ ] Update `transfer` to correct use the new policy fields. [small];
    - [ ] Update `preview transaction` to show the correct gas cost. [small];
    - [ ] Update `view transaction` to show the correct gas cost. [small];
    - [ ] Create a new `Policies` summary component to show the policies set to the transaction; [small];
      - [ ] Add component on the `Transaction view`;
      - [ ] Add component on the `Transaction Preview`;
    - [ ] Update all API changes that will be updated on the fuels-ts. [medium] (this needs a better understanding of what APIs will be changed and how they will be implemented);
 - Fuel Bridge
    - [ ]  Update all API changes that will be updated on the fuels-ts. [medium] (this needs a better understanding of what APIs will be changed and how they will be implemented);
 - Fuel Explorer
    -  [ ] Ensure the gas usage is showing the correct amount [small];
    - [ ] Create a new UI for showing policies. [small].

 small - up to one day
 medium - up to three days

Voxelot:
@luizstacio could this time be shortened by using fuel-tx wasm for serialization? cc @digorithm

luizstacio:
> @luizstacio could this time be shortened by using fuel-tx wasm for serialization? cc @digorithm

I think if we change fuel-tx this will be a big push on the fuels-ts side and potentially include multiple break changes.

src/tx-format/transaction.md
Voxelot: I just made gas limit a default field for script transactions. The only reason to keep it as a policy is to save a u64 off of basic transfers that don't use any script logic.

src/fuel-vm/instruction-set.md
```diff
@@ -2439,86 +2439,87 @@ Set `$rA` to the index of the currently-verifying predicate.
 
 Get [fields from the transaction](../tx-format/transaction.md).
 
-| name                                      | `imm`   | set `$rA` to                                     |
-|-------------------------------------------|---------|--------------------------------------------------|
-| `GTF_TYPE`                                | `0x001` | `tx.type`                                        |
-| `GTF_SCRIPT_GAS_PRICE`                    | `0x002` | `tx.gasPrice`                                    |
-| `GTF_SCRIPT_GAS_LIMIT`                    | `0x003` | `tx.gasLimit`                                    |
```
xgreenx, line 2446: We need to return back `GTF_SCRIPT_GAS_LIMIT`

xgreenx (approved):

PR #516, opened 2023-09-11
docs: polish "balanceOfStart" explanation

PaulRBerg:
I had to read the current phrase a couple of times to understand it.

Alternatively, we could say "There is given a" or "It is given a".


Voxelot (approved):
lgtm - thank you!

PR #518, opened 2023-09-28
Minor simplification to `Mint` transaction after the implementation

xgreenx:
Implementation: https://github.com/FuelLabs/fuel-vm/pull/593

### DIFF
--- a/src/fuel-vm/index.md
+++ b/src/fuel-vm/index.md
@@ -172,7 +172,7 @@ A call frame consists of the following, word-aligned:
 | 32    | `byte[32]`    | `to`       | Contract ID for this call.                                                    |
 | 32    | `byte[32]`    | `asset_id` | asset ID of forwarded coins.                                                  |
 | 8*64  | `byte[8][64]` | `regs`     | Saved registers from previous context.                                        |
-| 8     | `uint16`      | `codesize` | Code size in bytes, padded to word alignment.                                 |
+| 8     | `uint64`      | `codesize` | Code size in bytes, padded to word alignment.                                 |
 | 8     | `byte[8]`     | `param1`   | First parameter.                                                              |
 | 8     | `byte[8]`     | `param2`   | Second parameter.                                                             |
 | 1*    | `byte[]`      | `code`     | Zero-padded to 8-byte alignment, but individual instructions are not aligned. |
--- a/src/tx-format/transaction.md
+++ b/src/tx-format/transaction.md
@@ -133,16 +133,15 @@ Creates a contract with contract ID as computed [here](../identifiers/contract-i
 The transaction is created by the block producer and is not signed. Since it is not usable outside of block creation or execution, all fields must be fully set upon creation without any zeroing.
 This means that the transaction ID must also include the correct `txPointer` value, not zeroed out.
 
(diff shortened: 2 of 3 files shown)


src/tx-format/transaction.md
```diff
@@ -133,16 +133,15 @@ Creates a contract with contract ID as computed [here](../identifiers/contract-i
...
+| `outputContract` | [OutputContract](./output.md) | The contract utxo that assets are being minted to.   |
+| `mintAmount`     | `uint64`                      | The amount of funds minted.                          |
+| `mintAssetId`    | `byte[32]`                    | The asset IDs corresponding to the minted amount.    |
 
 Transaction is invalid if:
 
 - `txPointer` is zero or doesn't match the block.
-- `inputContract` is not of the type `InputType.Contract`
-- `outputContract` is not of the type `OutputType.Contract`
+- `outputContract.inputIndex` is not zero
```
bvrooman, line 147: > - `outputContract.inputIndex` is not zero

If I understand correctly, the mint transaction is effectively the coinbase. So it makes sense that the  has to be the first transaction. Is that what this change is saying?

> - `inputContract` is not of the type `InputType.Contract`
> - `outputContract` is not of the type `OutputType.Contract`

Are you removing these points because they are no longer true? Can the inputs and outputs be any type?

xgreenx: > If I understand correctly, the mint transaction is effectively the coinbase. So it makes sense that the has to be the first transaction. Is that what this change is saying?

Nope, each `OutputContract` should be connected to `InputContract`. Because we always have only one `InputContract` the `OutputContract.inputIndex` can be and should be only `0`=)

> It looks like we are constraining the inputs and outputs to Contract, so you are removing these points because there is no other option.

Yep, instead of `Input` and `Output` we work with `InputContract` and `OutputContract` directly removing possibilities of other `Input`/`Output` variants.

src/fuel-vm/index.md
```diff
@@ -172,7 +172,7 @@ A call frame consists of the following, word-aligned:
 | 32    | `byte[32]`    | `to`       | Contract ID for this call.                                                    |
 | 32    | `byte[32]`    | `asset_id` | asset ID of forwarded coins.                                                  |
 | 8*64  | `byte[8][64]` | `regs`     | Saved registers from previous context.                                        |
-| 8     | `uint16`      | `codesize` | Code size in bytes, padded to word alignment.                                 |
+| 8     | `uint64`      | `codesize` | Code size in bytes, padded to word alignment.                                 |
```
Voxelot, line 175: It doesn't need to be "padded to word alignment." if it's already a word right?

bvrooman (dismissed):
LGTM. You can remove "padded to word alignment" as @Voxelot mentioned. I personally don't mind having it because it implies that the 64 bit padding is part of the call frame specification, rather than a random occurrence.

src/fuel-vm/index.md
xgreenx, line 175:
```suggestion
| 8     | `uint64`      | `codesize` | Code size in bytes.                                 |
```

bvrooman (approved):

PR #525, opened 2023-10-17
docs: Specify max transaction size

bvrooman:
Related issues:
- Progresses https://github.com/FuelLabs/fuel-vm/issues/384

This PR specifies the maximum size of a transaction and that a transaction that exceeds the maximum size is invalid. 

This PR also renames the `constants.md` file to `consensus_parameters.md` to better reflect how these values are used. Since these values are not constants outlined by the specification, this change proposes to remove the `value` column altogether, allowing the current definition of consensus parameters to be decided by maintainers.


src/tx-format/constants.md
```diff
@@ -14,5 +14,6 @@
 | `MAX_SCRIPT_DATA_LENGTH`    | `uint64` |                 | Maximum length of script data, in bytes.      |
 | `MAX_MESSAGE_DATA_LENGTH`   | `uint16` |                 | Maximum length of message data, in bytes.     |
 | `MAX_STORAGE_SLOTS`         | `uint16` | `255`           | Maximum number of initial storage slots.      |
+| `MAX_TRANSACTION_SIZE`      | `uint64` | `17,000,000`    | Maximum size of a transaction, in bytes.      |
```
xgreenx, line 17: @Voxelot Do we want to have 17 MB or 18 MB? I don't remember=D

xgreenx (approved):

PR #531, opened 2023-11-04
Add BLS12-381 verification

SilentCicero:
# Abstract

Barreto-Lynn-Scott ("BLS") signature verificaiton will become more important for Fuel as we deploy to data constrained environments like Ethereum. In this PR we define a BLS signature verification operation for the Fuel Virtual Machine.

This operation allows two modes: singular and aggregate verification of BLS12-381 signatures.

## Notes
- Gas prices can be done based upon the number of messages to be aggregated upfront
 
Would request HashCloak to writeup the cryptographic specifics in the cryptography section of the specifications.

### DIFF
--- a/src/fuel-vm/instruction-set.md
+++ b/src/fuel-vm/instruction-set.md
@@ -110,6 +110,7 @@
   - [ECK1: Secp251k1 signature recovery](#eck1-secp256k1-signature-recovery)
   - [ECR1: Secp256r1 signature recovery](#ecr1-secp256r1-signature-recovery)
   - [ED19: edDSA curve25519 verification](#ed19-eddsa-curve25519-verification)
+  - [BLSA: Barreto-Lynn-Scott 12-381 signature verification](#blsa-bls12-381-verify)
   - [K256: keccak-256](#k256-keccak-256)
   - [S256: SHA-2-256](#s256-sha-2-256)
 - [Other Instructions](#other-instructions)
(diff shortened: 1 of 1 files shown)


Viktor110:
> # Abstract
> 
> Barreto-Lynn-Scott ("BLS") signature verification will become more important for Fuel as we deploy to data constrained environments like Ethereum. In this PR we define a BLS signature verification operation for the Fuel Virtual Machine.
> 
> 
> 
> This operation allows three modes: singular, aggregate and special aggregate verification of BLS12-381 signatures.
> 
> 
> 
> ## Proposed Modes
> 
> 1. **Single signature verification** - signature, message and public key are in memory
> 
> 2. **Aggregate verification** - signatures, messages, and public keys are in memory
> 
> 3. **Special aggregate verification** - (i.e. "allow the block producer to optionally aggregate my signature mode") - public key and message is in memory and is either included in an aggregate signature batch with other public keys, messages and a valid aggregate signature within the block or the individual signature is in memory
> 
> 
> 
> ## Notes
> 
> - Here for mode 3, the user would submit their transaction and witness data to the mempool, but there is no guarantee the signature will be available in memory as it may have been aggregated.
> 
> - Gas prices can be done based upon the number of messages to be aggregated upfront
> 
> - Perhaps the third option can fit into an optional transaction policy (as it's a special case)
> 
>  
> 
> Would request HashCloak to writeup the cryptographic specifics in the cryptography section of the specifications.

PR #534, opened 2023-11-07
Update Secp256r1 Instruction with Message Length as Immediate

SilentCicero:
## Abstract
The current specified implementation of the Secp256r1 opcode assumes a 32 byte message being signed over, however, in the current widespread use of the curve in WebAuthn and passkeys, the signed message is larger than 32 bytes (e.g. 69 bytes).

## Solution
Allow the immediate value to specify a more flexible message length beyond or below 32 bytes.

## Considerations
- For consistency, we should perhaps also enable a more flexible message length for the secp256k1 opcode.


Dentosal:
The thing we're signing with the opcode is not a message, but the hash of a message. In WebAuthn, the original message is condensed into the required 32 bytes using SHA-256, which is already available using `S256` instruction.

PR #535, opened 2023-11-09
Limit the number of receipts to u16::MAX

Dentosal:
Increases the receipt limit from `u8::MAX` to `u16::MAX`, leaving space for panic and script result. The old limit was not enforced, but https://github.com/FuelLabs/fuel-vm/pull/633 is going to enforce the new rules defined here.

The number of receipts is limited separtely from gas, since adding more receipts slows down all instructions that operate on them. Moreover, this change makes the upper bound for memory use clear.

### DIFF
--- a/src/identifiers/utxo-id.md
+++ b/src/identifiers/utxo-id.md
@@ -18,7 +18,7 @@ The ID of a message is computed as the [hash](../protocol/cryptographic-primitiv
 
 ### Message Nonce
 
-The nonce value for `InputMessage` is determined by the sending system and is published at the time the message is sent. The nonce value for `OutputMessage` is computed as the [hash](../protocol/cryptographic-primitives.md#hashing) of the [Transaction ID](./transaction-id.md) that emitted the message and the index of the message receipt `uint8`: `hash(byte[32] ++ uint8)`.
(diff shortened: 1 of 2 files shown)


src/identifiers/utxo-id.md
```diff
@@ -18,7 +18,7 @@ The ID of a message is computed as the [hash](../protocol/cryptographic-primitiv
 
 ### Message Nonce
 
-The nonce value for `InputMessage` is determined by the sending system and is published at the time the message is sent. The nonce value for `OutputMessage` is computed as the [hash](../protocol/cryptographic-primitives.md#hashing) of the [Transaction ID](./transaction-id.md) that emitted the message and the index of the message receipt `uint8`: `hash(byte[32] ++ uint8)`.
+The nonce value for `InputMessage` is determined by the sending system and is published at the time the message is sent. The nonce value for `OutputMessage` is computed as the [hash](../protocol/cryptographic-primitives.md#hashing) of the [Transaction ID](./transaction-id.md) that emitted the message and the index of the message receipt `uint16`: `hash(byte[32] ++ uint16)`.
```
xgreenx, line 21: Do we want to highlight the `uint16` uses canonical serialization?

Dentosal: Likely. The canonical serialization is just a `u64`, so it makes sense to highlight this.

Dentosal: I added that in 5a2dfe6

xgreenx (approved):

PR #545, opened 2024-02-13
Zero all malleable fields before execution

### LINKED ISSUE #542: Zero all malleable fields before execution

Dentosal:
Closes #542.

Also changes `tx.receiptsRoot` to only be updated after the tx has terminated, as the intermediate values are no longer useful. The speeds up and simplifies implementation of receipt-pushing opcodes.

### DIFF
--- a/src/fuel-vm/instruction-set.md
+++ b/src/fuel-vm/instruction-set.md
@@ -2497,11 +2497,6 @@ Get [fields from the transaction](../tx-format/transaction.md).
 | `GTF_INPUT_COIN_PREDICATE`                | `0x20B` | Memory address of `tx.inputs[$rB].predicate`                      |
 | `GTF_INPUT_COIN_PREDICATE_DATA`           | `0x20C` | Memory address of `tx.inputs[$rB].predicateData`                  |
 | `GTF_INPUT_COIN_PREDICATE_GAS_USED`       | `0x20D` | `tx.inputs[$rB].predicateGasUsed`                                 |
-| `GTF_INPUT_CONTRACT_TX_ID`                | `0x220` | Memory address of `tx.inputs[$rB].txID`                           |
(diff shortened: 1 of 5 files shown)


xgreenx:
Maybe it makes sense to remove the getter for malleable fields since they are zero all the time. We don't need to change the discriminant for values, we can leave gaps

xgreenx:
Before merging this PR, I would suggest doing https://github.com/FuelLabs/fuel-core/issues/1620 first. It will help identify all problematic fields. Maybe we missed something here

Dentosal:
CI not passing due to https://github.com/FuelLabs/fuel-specs/issues/547

src/fuel-vm/instruction-set.md
```diff
@@ -2491,19 +2490,13 @@ Get [fields from the transaction](../tx-format/transaction.md).
 | `GTF_INPUT_COIN_OWNER`                    | `0x203` | Memory address of `tx.inputs[$rB].owner`                          |
 | `GTF_INPUT_COIN_AMOUNT`                   | `0x204` | `tx.inputs[$rB].amount`                                           |
 | `GTF_INPUT_COIN_ASSET_ID`                 | `0x205` | Memory address of `tx.inputs[$rB].asset_id`                       |
-| `GTF_INPUT_COIN_TX_POINTER`               | `0x206` | Memory address of `tx.inputs[$rB].txPointer`                      |
```
Voxelot, line 2494: weren't we going to use tx_pointer as a replacement for the maturity field? Why are we not setting it anymore?

cc @xgreenx

Voxelot: Discussed on a call - the tx_pointer will be unreliable for predicates to use as a maturity check. The most sensible approach for now is to have predicates check the tx-wide maturity policy.

Voxelot (approved):

Voxelot:
For future reference, removing the intermediate receipt roots is possible now that we are going with a zk-hybrid proving approach.

xgreenx (approved):

PR #546, opened 2024-02-15
Update Specs to Include Gas Price changes

MitchTurner:
https://github.com/FuelLabs/fuel-specs/pull/546

This includes the changes for:
- Remove`gas_price` from tx
- Include `gas_price` on `Mint` tx
- Remove `GasPrice` policy
- Add `Tip` policy
- Make `MaxFee` mandatory

Since there is no concept of the client API in these specs, and also the implementation of tx checks, I didn't include:
- Anything wrt to `gas_price` estimate or latest `gas_price` API endpoints
- Anything wrt the metadata changes for `Checked<Tx>`

That's fine if we don't want the scope of the VM to include those sort of things, but it's not clear to me _what_ these specs are constrained by.

### DIFF
--- a/src/protocol/tx-validity.md
+++ b/src/protocol/tx-validity.md
@@ -238,7 +238,7 @@ def available_balance(tx, assetId) -> int:
 def unavailable_balance(tx, assetId) -> int:
     sentBalance = sum_outputs(tx, assetId)
     # Total fee balance
-    feeBalance = reserved_fee_balance(tx, assetId)
+    feeBalance = tx.policies.max_fee
     # Only base asset can be used to pay for gas
     if assetId == 0:
         return sentBalance + feeBalance
--- a/src/protocol/tx-validity.md
+++ b/src/protocol/tx-validity.md
@@ -214,19 +214,6 @@ def max_gas(tx) -> int:
     return gas
 
 
-def reserved_feeBalance(tx, assetId) -> int:
-    """
-    Computes the maximum potential amount of fees that may need to be charged to process a transaction.
-    """
-    maxGas = max_gas(tx)
-    feeBalance = gas_to_fee(maxGas, tx.gasPrice)
-    # Only base asset can be used to pay for gas
-    if assetId == 0:
-        return feeBalance
-    else:
-        return 0
-
-
 def available_balance(tx, assetId) -> int:
     """
     Make the data message balance available to the script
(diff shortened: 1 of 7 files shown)


src/tx-format/policy.md
```diff
@@ -14,11 +14,11 @@ enum PolicyType : uint32 {
...
 
-## `GasPrice`
+## `Tip`
 
-| name       | type     | description               |
-|------------|----------|---------------------------|
-| `gasPrice` | `uint64` | Gas price for transaction |
+| name       | type     | description                                                                                   |
+|------------|----------|-----------------------------------------------------------------------------------------------|
+| `gasPrice` | `uint64` | Additional, optional fee in `BASE_ASSET` to incentivize block producer to include transaction |
```
xgreenx, line 21:
```suggestion
| `tip` | `uint64` | Additional, optional fee in `BASE_ASSET` to incentivize block producer to include transaction |
```

src/protocol/tx-validity.md
```diff
@@ -212,14 +212,14 @@ def max_gas(tx) -> int:
     if tx.type == TransactionType.Script:
        gas = gas + tx.gasLimit
     return gas
-
-
-def reserved_feeBalance(tx, assetId) -> int:
```
MitchTurner, line 217: I couldn't find any function with this name in our code, and it sounds like `max_fee`, so I changed it. Might be wrong.

Dentosal (approved):

src/protocol/tx-validity.md
xgreenx: Yeah, in the code analog is `max_fee` function=) Could you also update other places here(`reserved_fee_balance`), please?

src/tx-format/policy.md
```diff
@@ -44,11 +44,12 @@ Transaction is invalid if:
...
-| name      | type     | description                                                     |
-|-----------|----------|-----------------------------------------------------------------|
-| `max_fee` | `uint64` | The maximum fee payable by this transaction using `BASE_ASSET`. |
+| name      | type     | description                                                                                                                                |
+|-----------|----------|--------------------------------------------------------------------------------------------------------------------------------------------|
+| `max_fee` | `uint64` | The maximum fee payable by this transaction using `BASE_ASSET`. This is used to check transactions before the actual `gas_price` is known. |
 
 Transaction is invalid if:
 
+- `max_fee` is not set
```
xgreenx, line 53: Maybe it is better to move this check to where the gas price was.

xgreenx (changes requested):
You also need to update `unavailable_balance` function to use `max_fee` policy instead of the `maxFee`(`reserved_fee_balance`) function

src/tx-format/policy.md
MitchTurner: I'm not sure I understand. Move this section to where the `## GasPrice` section was? Which "gas price" are you referring to?

Issue #553, opened 2024-03-03
Consider increasing maximum size of inputs, outputs and witnesses to be 2 bytes


xgreenx:
The comment from @SilentCicero:
> Changing the byte size of the inputs, outputs and witness lengths from 1 byte (max 255 value) to 2 bytes to enable values of up to 65025. I believe this would be incredibly useful for cases where block producers want to clear a lot of predicate based state changes over a single transaction. This deeply impacts how effective things like the signature aggregation are, as those opcodes likely only work in single transactional environments.

Concerns that we need to think about before implementing it:
- Increasing them to 2 bytes may impact the performance in the worst scenario. First of all our codebase assume in some places that it can't be more that `255` and uses non-optimal implementation.
- It may affect the way how we benchmark opcodes. As an example, right now, we assume that there couldn't be more than 255 receipts generated by the transaction because of the limit on inputs.
- It may break some existing tooling and maybe even contracts because they assume that the actual transaction lives after 255 assets.
    <img width="608" alt="image" src="https://github.com/FuelLabs/fuel-specs/assets/18346821/012ae83b-c161-477a-b195-be7cd3f4a5a7">
- From the formula above we can calculate how much space we need to reserve in the worst scenario: `32 + 8 + 2^16 * 40 = 2,621,480` ~ 2.5 MB. It is too much to do for each predicate/script and we need to add support for dynamic `tx_offset`. 
- It is a very good vector for DoS attacks since all `2^16 -1` inputs may be valid, while the last one is invalid. And the node should verify all of them before discarding the transaction.

SilentCicero:
My recommendation would be to increase to 2 bytes in the transaction format, but enforce a VM limit constraint that we can upgrade for later. For now at a max value of 255.

This way, all existing execution assumptions remain the same around the value, and we can do an update later to lift the limit once we believe it's safe to do so.

Thoughts?

xgreenx:
Yeah, that sound good. We need to change the `witnes_index` and `input_index` to be `u16` as well.

PR #558, opened 2024-03-13
Increase size of inputs, outputs and witnesses to uint16

Dentosal:
Closes https://github.com/FuelLabs/fuel-specs/issues/553

VM PR: https://github.com/FuelLabs/fuel-vm/pull/698


src/fuel-vm/index.md
```diff
@@ -25,6 +25,9 @@ This document provides the specification for the Fuel Virtual Machine (FuelVM).
 | `MEM_MAX_ACCESS_SIZE`   | `uint64` |         | Maximum memory access size, in bytes.   |
 | `VM_MAX_RAM`            | `uint64` | `2**26` | 64 MiB.                                 |
 | `MESSAGE_MAX_DATA_SIZE` | `uint64` |         | Maximum size of message data, in bytes. |
+| `MAX_INPUTS`            | `uint16` | `256`   | Maximum number of inputs.               |
```
xgreenx, line 28: Hmm, we are using consensus parameters inside of the VM. And we have these fields there https://github.com/FuelLabs/fuel-specs/blob/master/src/tx-format/consensus_parameters.md

Maybe we need to use them here instead

Dentosal: True. Removed these in 1227f77

xgreenx (approved):

PR #562, opened 2024-03-22
Grouped unique fields for transactions at the beginning

xgreenx:
Soon, we will add several more transaction variants that are also chargeable. Because of that, I've decided to move unique fields of the transaction to the top while keeping the common part at the end.

Implementation part with description of how it can be used https://github.com/FuelLabs/fuel-vm/pull/703#issue-2202505020

### DIFF
--- a/src/tx-format/transaction.md
+++ b/src/tx-format/transaction.md
@@ -58,13 +58,13 @@ enum ReceiptType : uint8 {
 | name               | type                        | description                             |
 |--------------------|-----------------------------|-----------------------------------------|
 | `scriptGasLimit`   | `uint64`                    | Gas limits the script execution.        |
+| `receiptsRoot`     | `byte[32]`                  | Merkle root of receipts.                |
 | `scriptLength`     | `uint64`                    | Script length, in instructions.         |
 | `scriptDataLength` | `uint64`                    | Length of script input data, in bytes.  |
 | `policyTypes`      | `uint32`                    | Bitfield of used policy types.          |
 | `inputsCount`      | `uint16`                    | Number of inputs.                       |
 | `outputsCount`     | `uint16`                    | Number of outputs.                      |
 | `witnessesCount`   | `uint16`                    | Number of witnesses.                    |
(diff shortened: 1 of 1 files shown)


Dentosal (approved):

src/tx-format/transaction.md
```diff
@@ -104,14 +104,14 @@ The receipts root `receiptsRoot` is the root of the [binary Merkle tree](../prot
...
-| `policyTypes`          | `uint32`                    | Bitfield of used policy types.                    |
+| `salt`                 | `byte[32]`                  | Salt.                                             |
 | `storageSlotsCount`    | `uint64`                    | Number of storage slots to initialize.            |
+| `policyTypes`          | `uint32`                    | Bitfield of used policy types.                    |
 | `inputsCount`          | `uint16`                    | Number of inputs.                                 |
 | `outputsCount`         | `uint16`                    | Number of outputs.                                |
 | `witnessesCount`       | `uint16`                    | Number of witnesses.                              |
-| `salt`                 | `byte[32]`                  | Salt.                                             |
-| `policies`             | [Policy](./policy.md)`[]`   | List of policies.                                 |
 | `storageSlots`         | `(byte[32], byte[32]])[]`   | List of storage slots to initialize (key, value). |
```
MitchTurner, line 113: I'm kinda confused by how we are specifying the fields of `TransactionCreate` It looks like the first 8 here match what we have in `fuel-vm`, but with `witnessCount` instead of `witness`. But here we also have 5 more that are duplicates.

I guess I'm trying to understand why we have `inputsCount` AND `inputs`, etc, in the specs, but just `inputs` in the code.

xgreenx: In the canonical serialization and deserialization, we have a static parts and a dynamic parts. The static part is always the same regardless of the content. In the case of transactions, you always know that unique fields are static, and that is why they are at the top. The dynamic data always mostly lives at the end.

That schema allows us to get the main information about the transaction without deserializing the whole transaction.

An example of the `canonical::Serialize` trait definition:

image

PR #567, opened 2024-04-05
Added a new `Upgrade` transaction to perform a network upgrade

xgreenx:
Corresponding implementation: https://github.com/FuelLabs/fuel-vm/pull/707

The change adds a new `Upgrade` transaction that allows upgrading either consensus parameters or state transition function used by the network to produce future blocks.

The purpose of the upgrade is defined by the `Upgrade Purpose` type.

The `Upgrade` transaction is chargeable, and the sender should pay for it. Transaction inputs should contain only base assets.

Only the privileged address can upgrade the network. The privileged address can be either a real account or a predicate.

We use postcard algorithm to serialize and deserialize consensus parameters during the upgrade. This algorithm works with numbers in a more sufficient way and compresses them well.


src/protocol/tx-validity.md
```diff
@@ -179,6 +179,12 @@ def metadata_gas_fees(tx) -> int:
     elif tx.type == TransactionType.Script:
         # add intrinsic cost of calculating the transaction id
         total += sha256_gas_fee(size(tx))
+    elif tx.type == TransactionType.Upgrade:
+        # add intrinsic cost of calculating the transaction id
+        total += sha256_gas_fee(size(tx))
```
Dentosal, line 184: Isn't this same for all tx types? Could we maybe move it outside the if statement, e.g. to the initializer of `total`? Also above we add it for every created output contract which doesn't seem correct. Not really part of this PR, but noticed that while reviewing.

src/tx-format/consensus_parameters.md
```diff
@@ -1,20 +1,21 @@
...
-| `MAX_PREDICATE_DATA_LENGTH` | `uint64`  | Maximum length of predicate data, in bytes.   |
-| `MAX_SCRIPT_LENGTH`         | `uint64`  | Maximum length of script, in instructions.    |
-| `MAX_SCRIPT_DATA_LENGTH`    | `uint64`  | Maximum length of script data, in bytes.      |
-| `MAX_MESSAGE_DATA_LENGTH`   | `uint64`  | Maximum length of message data, in bytes.     |
-| `MAX_STORAGE_SLOTS`         | `uint64`  | Maximum number of initial storage slots.      |
-| `MAX_TRANSACTION_SIZE`      | `uint64`  | Maximum size of a transaction, in bytes.      |
-| `MAX_WITNESSES`             | `uint64`  | Maximum number of witnesses.                  |
-| `CHAIN_ID`                  | `uint64`  | A unique per-chain identifier.                |
-| `BASE_ASSET_ID`             | `bytes32` | The base asset of the chain.                  |
+| name                        | type      | description                                                    |
```
Dentosal, line 3: meta: markdown tables don't require leading and trailing `|` on tables, and removing them would require us to include the whole table in diff less often

src/tx-format/transaction.md
```diff
@@ -123,9 +124,9 @@ Given helper `sum_variants()` that sums all variants of an enum.
 Transaction is invalid if:
 
 - Any input is of type `InputType.Contract` or `InputType.Message` where `input.dataLength > 0`
+- Any input uses non-base asset.
 - Any output is of type `OutputType.Contract` or `OutputType.Variable` or `OutputType.Message`
-- More than one output is of type `OutputType.Change` with `asset_id` of zero
-- Any output is of type `OutputType.Change` with non-zero `asset_id`
+- Any output is of type `OutputType.Change` with non-base `asset_id`
```
Dentosal, line 129: nit: Out-of-scope for this PR. Still nice fixes and having them here is ok for me.

src/tx-format/transaction.md
```diff
@@ -159,3 +160,36 @@ Transaction is invalid if:
...
+| `witnessesCount` | `uint16`                               | Number of witnesses.           |
+| `policies`       | [Policy](./policy.md)`[]`              | List of policies.              |
+| `inputs`         | [Input](./input.md)`[]`                | List of inputs.                |
+| `outputs`        | [Output](./output.md)`[]`              | List of outputs.               |
+| `witnesses`      | [Witness](./witness.md)`[]`            | List of witnesses.             |
+
+Given helper `max_gas()` returns the maximum gas that the transaction can use.
+Given helper `count_ones()` that returns the number of ones in the binary representation of a field.
+Given helper `count_variants()` that returns the number of variants in an enum.
+Given helper `sum_variants()` that sums all variants of an enum.
```
Dentosal, line 181: We're repeating these quite a bit. Maybe we should have a centralized list of these? Although I think these might be almost evident enough tha they don't need explanations.

src/tx-format/upgrade_purpose.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+| name   | type                                                                                        | description              |
+|--------|---------------------------------------------------------------------------------------------|--------------------------|
+| `type` | [`UpgradePurposeType`](#upgradepurposetype)                                                 | Type of upgrade purpose. |
+| `data` | One of [`ConsensusParameters`](#consensusparameters), [`StateTransition`](#statetransition) | Upgrade purposes.        |
+
+Transaction is invalid if:
+
+- `type > UpgradePurposeType.StateTransition`
```
Dentosal, line 17: nit: I'd prefer to just say that "`type` is not valid `UpgradePurposeType` value"

spell-check-custom-words.txt
```diff
@@ -50,6 +51,8 @@ Sway
 TAI
 TODO
 TypeScript
+UpgradePurpose
```
sarahschwartz, line 54: Instead of adding UpgradePurpose and UpgradePurposeType to this file, you can just wrap them in backticks like `UpgradePurpose` so the spell checker will ignore them. Otherwise LGTM!

spell-check-custom-words.txt
```diff
@@ -50,6 +51,8 @@ Sway
...
+UpgradePurposeType
```
Dentosal: nit: these should probably just be in backticks (`code`) everywhere, so we don't need to spellcheck-allow them

Dentosal (changes requested):
This describes the new data types and validity rules. However, I think the semantics of an upgrade should be described as well.

src/tx-format/consensus_parameters.md
xgreenx: My formatter throws an error if I don't use `|`=)

Dentosal: Github or any renderer really wont mind but ok we can keep them

Dentosal (approved):
Now I'm happy with this

src/protocol/tx-validity.md
```diff
@@ -174,9 +174,12 @@ def metadata_gas_fees(tx) -> int:
                 # add intrinsic cost of calculating the contract id 
                 # size = 4 byte seed + 32 byte salt + 32 byte code root + 32 byte state root
                 total += sha256_gas_fee(100)
-                # add intrinsic cost of calculating the transaction id
-                total += sha256_gas_fee(size(tx))
-    elif tx.type == TransactionType.Script:
```
MitchTurner, line 179: Wait, what happened to the `Script` path?

xgreenx: It is handled below=) 
image

Since that part is identical for all transactions except `Mint`

PR #570, opened 2024-04-09
A new `Upload` transaction to upload the huge bytecode on the chain

xgreenx:
Corresponding implementation: https://github.com/FuelLabs/fuel-vm/pull/720
The change adds a new `Upload` transaction that allows uploading huge byte code on chain subsection by subsection.

The `Upload` transaction is chargeable and is twice as expensive as the `Create` transaction. Anyone can submit this transaction.

The specification contains more description about the flow.

### DIFF
--- a/src/protocol/tx-validity.md
+++ b/src/protocol/tx-validity.md
@@ -178,6 +178,11 @@ def metadata_gas_fees(tx) -> int:
         if tx.upgradePurpose.type == UpgradePurposeType.ConsensusParameters:
             # add intrinsic cost of calculating the consensus parameters hash
             total += sha256_gas_fee(size(tx.witnesses[tx.upgradePurpose.witnessIndex].data))
+    elif tx.type == TransactionType.Upload:
+        # add intrinsic cost of calculating the root based on the number of bytecode subsections
+        total += contract_state_root_gas_fee(tx.subsectionsNumber)
+        # add intrinsic cost of hashing the subsection for verification of the connection with Binary Merkle tree root
+        total += sha256_gas_fee(size(tx.witnesses[tx.witnessIndex]))
             
     if tx.type != TransactionType.Mint:
         # add intrinsic cost of calculating the transaction id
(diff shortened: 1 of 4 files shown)


src/protocol/tx-validity.md
```diff
@@ -178,6 +178,11 @@ def metadata_gas_fees(tx) -> int:
         if tx.upgradePurpose.type == UpgradePurposeType.ConsensusParameters:
             # add intrinsic cost of calculating the consensus parameters hash
             total += sha256_gas_fee(size(tx.witnesses[tx.upgradePurpose.witnessIndex].data))
+    elif tx.type == TransactionType.Upload:
+        # add intrinsic cost of calculating the root based on the number of bytecode subsections
+        total += contract_state_root_gas_fee(tx.subsectionsNumber)
```
Dentosal, line 183: Is this function explained somewhere? What does it do, exactly? The MT grows `log(subsections)`, right?

src/tx-format/transaction.md
```diff
@@ -193,3 +194,40 @@ Transaction is invalid if:
...
+| `inputsCount`       | `uint16`                    | Number of inputs.                                                                       |
+| `outputsCount`      | `uint16`                    | Number of outputs.                                                                      |
+| `witnessesCount`    | `uint16`                    | Number of witnesses.                                                                    |
+| `proofSet`          | `byte[32][]`                | The proof set of Merkle nodes to verify the connection of the subsection to the `root`. |
+| `policies`          | [Policy](./policy.md)`[]`   | List of policies.                                                                       |
+| `inputs`            | [Input](./input.md)`[]`     | List of inputs.                                                                         |
+| `outputs`           | [Output](./output.md)`[]`   | List of outputs.                                                                        |
+| `witnesses`         | [Witness](./witness.md)`[]` | List of witnesses.                                                                      |
+
+Transaction is invalid if:
```
Dentosal, line 225: If I understand correctly, we also need this:

> - `subsectionIndex >= subsectionsNumber`

Also, is there a way to verify that subsectionsNumber doesn't change between Upload txs?

src/protocol/tx-validity.md
xgreenx: We don't have an explanation for this function. Basically, all functions that are used here are based on our benchmark.

src/tx-format/transaction.md
xgreenx: `subsectionIndex >= subsectionsNumber` is enforced by the root verification, but I added it just for safety=)
The verification of the proof should guarantee the same `subsectionsNumber`. Changing `subsectionsNumber` affects the Merkle tree path.

Right now we have a bug in the `verify` implementation: https://github.com/FuelLabs/fuel-vm/issues/716

Dentosal (approved):

PR #585, opened 2024-06-03
Zero $flag on call

Dentosal:
VM PR: https://github.com/FuelLabs/fuel-vm/pull/743

Called contracts shouldn't inherit the `$flag` register.

### Before requesting review


xgreenx (approved):

xgreenx:
Do we need to mention that during return back, we are restoring the flag?

Dentosal:
> Do we need to mention that during return back, we are restoring the flag?

We already mention that, as restore was the old behavior as well.

Issue #589, opened 2024-06-20
Blob tx and loading


Dentosal:
Tracking issue for `Blob` transactions and related opcodes.

As discussed with the client team on 2024-06-20, to allow contracts larger than the contract size limit, and to allow upgradability, we'll be adding a new tx type that's just a bunch of bytes, and opcodes to load subsections of that. The whole thing reminds contracts and `LDC` a bit, but blob won't have state or balances, and the corresponding load instructions will be slightly different.

Implemented in:
- https://github.com/FuelLabs/fuel-specs/pull/592
- https://github.com/FuelLabs/fuel-vm/pull/780
- https://github.com/FuelLabs/fuel-core/pull/1988

xgreenx:
All sub-tasks are done=)

PR #592, opened 2024-06-24
Blob tx and instructions

Dentosal:
Work towards https://github.com/FuelLabs/fuel-specs/issues/589. The VM PR https://github.com/FuelLabs/fuel-vm/pull/780 contains a working implementation.

To allow contracts larger than the contract size limit, and to allow upgradability, we'll be adding a new tx type that's just a bunch of bytes, and opcodes to load subsections of that. A *blob* will be like a contract, but without state or balances, allowing it to be quite a bit cheaper. The new instructions `BSIZ` and `BLDD` and the new blob mode for `LDC` closely follow how `CSIZ`, `LDC` and `CCP` work, but only operation on blob data.

### Before requesting review


IGI-111:
I know it's bikeshedding but can we call it something other than "blob"? It evokes either object based storage or proprietary binaries in kernel land. What this really is is an executable.

Dentosal:
> I know it's bikeshedding but can we call it something other than "blob"? It evokes either object based storage or proprietary binaries in kernel land. What this really is is an executable.

From vm and storage layer perspective, this is just an opaque bytearray, addressed by it's hash. The executability is just a nice side effect of the generic design, although it's also the main purpose for introducing this.

IGI-111:
> it's also the main purpose for introducing this

That's my point really. Clarity of intent. I understand that this is just a distributed byte array. But it's not meant to store things, it's meant to store code that will be executed.

"Executable" might not be a good name for it since it is not guaranteed to be a whole executable.

Really it's just a bunch of loadable bytecode.

segfault-magnet (commented):
LDC docs should be updated to show `mode` and that the first arg can now be a blob_id.

Voxelot:
@Dentosal should this PR also include the changes to the LDC opcode?

Dentosal:
> @Dentosal should this PR also include the changes to the LDC opcode?

Yes. They were not originally included as it was unclear whether we would have them in LDC or a separate instruction, but it seems like we have converged on LDC. The new changes are included in 71f7462.

PR #596, opened 2024-07-01
Updates ABI spec with hash based ids, versions and program type.

esdrubal:
[#5151](https://github.com/FuelLabs/sway/issues/5151)
[#5952](https://github.com/FuelLabs/sway/issues/5952)
[#5954](https://github.com/FuelLabs/sway/issues/5954)

Adds specification for how hashes based ids are generated from a string representation of types.

Updates `json-abi-format.md` with versions, program type, and hash based ids.

### Before requesting review

- [ ] [Typescript SDK](https://github.com/FuelLabs/fuels-ts/)
- [ ] Someone else?

### DIFF
--- a/src/abi/hash-based-ids.md
+++ b/src/abi/hash-based-ids.md
@@ -1,0 +1,72 @@
+# Hash based IDs
+
+Hash based ids are deterministically generated from associated types and are used in the JSON ABI for `type` IDs and for `logId`.
+This document specifies how the hash based IDS are generated for `type` IDs and for `logId`.
+
+## Generation
+
+Hash based ids for `type` IDs are generated from the `sha256` of a string that represents the type.
+
+For `logIds` we use the first 8 bytes of the `sha256` of a string that represents the type, this is because the `LOG` and `LOGD` opcodes use a 64bit value as log id.
+
+## String representation of types
+
+For describing the string representation of type we will use the notation `{abi_str(T)}` that should be replaced by the respective ABI string representation of the respective type `T`.
+
+### Intrinsics
+
+ `u8`   => `"u8"`
+ `u16`  => `"u16"`
+ `u32`  => `"u32"`
+ `u64`  => `"u64"`
+ `u256` => `"u256"`
+ `b256` => `"b256"`
+ `bool` => `"bool"`
+
+### String arrays
+
+  String array of size `1` => `"str[1]"`
+  String array of size `2` => `"str[2]"`
+  etc.
+
+### String slices
+
+  String slice => `"str"`
+
+### Arrays
+
+ `[T; 1]` => `"[{abi_str(T)}; 1]"`
+ `[T; 2]` => `"[{abi_str(T)}; 2]"`
+etc.
+
+### Tuples
+
+ `()`      => `"()"`
+ `(T1)`    => `"({abi_str(T1)})"`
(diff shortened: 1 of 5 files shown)


src/abi/json-abi-format.md
```diff
@@ -26,51 +32,54 @@ we define the following expressions:
 
 The ABI of a contract is represented as a JSON object containing the following properties:
 
-- `"types`": an array describing all the _type declarations_ used (or transitively used) in the ABI. Each _type declaration_ is a JSON object that contains the following properties:
-  - `"typeId"`: a unique integer ID.
+- `"specVersion"`: a string representing the version pointing to this document versioning. When `specVersion` is incremented, it will usually incur an increment in `abiVersion` too.
```
arboleya, line 35: Is `specVersion` replacing the `encoding` property?

encoding

src/abi/json-abi-format.md
```diff
@@ -26,51 +32,54 @@ we define the following expressions:
...
+- `"abiVersion"`: a string representing the version of the implementation of this document `abiVersion` may be incremented while `specVersion` is not.
```
arboleya, line 36: In which situation should one care for `abiVersion`?

src/abi/hash-based-ids.md
```diff
@@ -0,0 +1,72 @@
...
+### Tuples
+
+ `()`      => `"()"`
+ `(T1)`    => `"({abi_str(T1)})"`
+ `(T1,T2)` => `"({abi_str(T1)}, {abi_str(T2)})"`
+etc.
+
+### Enums
+
+  `Option` enum with type parameter `T` => `"enum std::option::Option<{abi_str(T)}>"`
```
nedsalk, line 51: Does this mean that the `--json-abi-with-callpaths` is now on by default?

src/abi/json-abi-format.md
IGI-111: We could have breaking changes in the spec without breaking chances in the ABI and vice versa.

Also, a new contract may want to deliberately use an old ABI version for compatibility reasons.

I really think these should be different.

arboleya: So, if the spec is at version `3` but users deliberately chose to use version `1`:

```json
{
  "specVersion": "3",
  "abiVersion": "1"
}
```

arboleya: @nedsalk @hal3e Did you understand the distinction between `specVersion` and `abiVersion`?

I suspect I might still be misinterpreting it.

hal3e: @IGI-111 can you please give us an example of a breaking change in the spec that does not break the abi and vice versa ?

arboleya: > Something that would bump the abiVersion would be for instance changing how intrinsic types are encoded

In this case, what's the difference between `abiVersion` and `encoding`?

IGI-111: If we already have an encoding field, we don't need an abiVersion one I think.

arboleya: @esdrubal @IGI-111 It is a NIT, but should we use the same suffix for both?
- `specVersion`
- `encodingVersion`

Renaming `encoding` -> `encodingVersion` is pretty easy for us.

arboleya: Thank you.

arboleya (approved):

IGI-111 (approved):

PR #600, opened 2024-07-24
change ed19 verify operator, to comply with ed spec

luizstacio:
When using EdDSA (Ed25519) to verify a signature, the payload is not limited by 32 bytes.

- [ ] [Connectors](https://github.com/FuelLabs/fuel-connectors/)

### DIFF
--- a/src/fuel-vm/instruction-set.md
+++ b/src/fuel-vm/instruction-set.md
@@ -2286,7 +2286,7 @@ Panic if:
 
 Verification are specified [here](../protocol/cryptographic-primitives.md#eddsa-public-key-cryptography).
 
-If register `$rD` is provided with `0` or not provided. It will be set to default `32`.
+If register `$rD` is provided with `0`. It will be set to default `32`.
 
 If there is an error in verification, `$err` is set to `1`, otherwise `$err` is cleared.
 
--- a/src/fuel-vm/instruction-set.md
+++ b/src/fuel-vm/instruction-set.md
@@ -2275,7 +2275,7 @@ To get the address from the public key, hash the public key with [SHA-2-256](../
 | Operation   | ```ed19verify(MEM[$rA, 32], MEM[$rB, 64], MEM[$rC, $rD]);```                                                                |
 | Syntax      | `ed19 $rA, $rB, $rC, $rD`                                                                                                   |
 | Encoding    | `0x00 rA rB rC rD`                                                                                                          |
-| Notes       | Takes message instead of hash. **For backwards compatibility reasons, if `$rD == 0`, it will be set treated as `32`.**      |
+| Notes       | Takes message instead of hash. **For backwards compatibility reasons, if `$rD == 0`, it will be treated as `32`.**          |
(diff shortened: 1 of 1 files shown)


Dentosal:
This seems completely incorrect. The ED19 "payload" is a message **hash**, typically sha-2-256. Using the instruction with a non-hashed value is cryptographically unsound. The payload is not hashed again by the instruction. To use `ED19` with any message of any size, you must first hash it, typically using `S256` instruction to do so.

luizstacio:
> Ed25519

https://en.wikipedia.org/wiki/EdDSA on Signature verification

$$
H(R \parallel A \parallel M) A
$$

From my understanding the message is required by the algo to be able to perform the hash on the validation.
Forcing to hash the message is not an option as you can't control the environment where the signature is going to happen, ex.: Wallets.

src/fuel-vm/instruction-set.md
```diff
@@ -2269,19 +2269,19 @@ To get the address from the public key, hash the public key with [SHA-2-256](../
...
-| Notes       |                                                                                                                                                     |
+|             |                                                                                                                             |
+|-------------|-----------------------------------------------------------------------------------------------------------------------------|
+| Description | Verification 64-byte signature at `$rB` with 32-byte public key at `$rA` for a message starting at `$rC` with length `$rD`. |
+| Operation   | ```ed19verify(MEM[$rA, 32], MEM[$rB, 64], MEM[$rC, $rD]);```                                                                |
+| Syntax      | `ed19 $rA, $rB, $rC, $rD`                                                                                                   |
+| Encoding    | `0x00 rA rB rC rD`                                                                                                          |
+| Notes       | Takes message instead of hash. **For backwards compatibility reasons, if `$rD == 0`, it will be set treated as `32`.**      |
```
mchristopher, line 2278: Minor, but can we say "the value of `$rD` will default to `32`" instead of "it will be set treated as `32`"

Dentosal: "set treated" was typo, definitely not good. However, saying that it "will default to 32" isn't technically correct, as zero would be a valid value as well if not for this mapping. (i.e. you can say that a missing value defaults to 32, but since the value cannot be missing, it's weird to say that). 

I changed it to be just

> if `$rD == 0`, it will be treated as `32`

But maybe something like

> `if $rD == 0` will default to `32` instead

could be used as well

mchristopher: But `null` is also valid here if the dev is using an old version of the opcode, so we should say "if `$rD == 0` or is `null`, the value of `$rD` will default to `32`", or something in that realm, correct?

Dentosal: There's no "null" in the instruction representation. The field is just a six-bit integer, and the old version required those bits to be zeroes.

mchristopher: Ah, got it. Sounds good to me then, thanks!

mchristopher (approved):

PR #611, opened 2024-10-06
Add LDC mode 2, which allows code from memory

Dentosal:
VM PR https://github.com/FuelLabs/fuel-vm/pull/849

### Before requesting review

### DIFF
--- a/src/fuel-vm/instruction-set.md
+++ b/src/fuel-vm/instruction-set.md
@@ -1804,7 +1804,7 @@ Panic if:
 | Operation   | `code = match imm { 0 => contract_code(mem[$rA,32]), 1 => blob_payload(mem[$rA,32]), 2 => mem[$ra, ..] }; MEM[$ssp, $rC] = code[$rB, $rC];`       |
 | Syntax      | `ldc $rA, $rB, $rC, imm`                                                                                                                          |
 | Encoding    | `0x00 rA rB rC imm`                                                                                                                               |
-| Notes       | If `$rC` is greater than the code size, zero bytes are filled in.                                                                                 |
+| Notes       | If `$rC` is greater than the code size, zero bytes are filled in. Length final is always padded to word boundary.                                 |
 
 Object type from `imm` determines the source for loading as follows:
 
--- a/src/fuel-vm/instruction-set.md
+++ b/src/fuel-vm/instruction-set.md
@@ -91,7 +91,7 @@
   - [`CCP`: Code copy](#ccp-code-copy)
   - [`CROO`: Code Merkle root](#croo-code-merkle-root)
   - [`CSIZ`: Code size](#csiz-code-size)
-  - [`LDC`: Load code from an external contract](#ldc-load-code-from-an-external-contract-or-blob)
+  - [`LDC`: Load code from an external contract, blob or memory](#ldc-load-code-from-an-external-contract-blob-or-memory)
   - [`LOG`: Log event](#log-log-event)
   - [`LOGD`: Log data event](#logd-log-data-event)
   - [`MINT`: Mint new coins](#mint-mint-new-coins)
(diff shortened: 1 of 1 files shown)


src/fuel-vm/instruction-set.md
```diff
@@ -1796,33 +1796,35 @@ Panic if:
...
 | other | _reserved_    |
 
 Panic if:
 
 - `$ssp + $rC` overflows or `> VM_MAX_RAM`
-- `$rA + 32` overflows or `> VM_MAX_RAM`
+- `imm <= 1` and `$rA + 32` overflows or `> VM_MAX_RAM`
 - `$ssp + $rC >= $hp`
 - `imm == 0` and `$rC > CONTRACT_MAX_SIZE`
 - `imm == 0` and contract with ID `MEM[$rA, 32]` is not in `tx.inputs`
```
xgreenx, line 1824: Do we want to hi light that we return an error if it is called in the context of predicates?

Dentosal: It's already called out in the predicate section of https://github.com/FuelLabs/fuel-specs/pull/612, but we could do that.

Dentosal, line 1824:
```suggestion
- `imm == 0` and context is a predicate
```

Maybe like this?

src/fuel-vm/instruction-set.md
```diff
@@ -1796,33 +1796,35 @@ Panic if:
...
 - Contract with ID `MEM[$rB, 32]` is not in `tx.inputs`
 
-### `LDC`: Load code from an external contract or blob
+### `LDC`: Load code from an external contract, blob or memory
 
 |             |                                                                                                                                                   |
 |-------------|---------------------------------------------------------------------------------------------------------------------------------------------------|
-| Description | Copy `$rC` bytes of code at offset `$rB` from object with 32 byte id starting at `$rA` into memory starting at `$ssp`. Object type is in `imm`.   |
-| Operation   | `id = mem[$rA,32]; code = match imm { 0 => contract_code($id), 1 => blob_payload($id) }; MEM[$ssp, $rC] = code[$rB, $rC];`                        |
+| Description | Copy `$rC` bytes of code at offset `$rB` from object identified with `$rA` into memory starting at `$ssp`. Object type is in `imm`.               |
```
acerone85, line 1803: nit: it would be nice to explain exactly what it is meant by "identified". E.g. something "with hash equivalent to the content of `$ra`.

Dentosal: But it's not always a hash, which is why I use an abstract wording here. The table below explains the types.

src/fuel-vm/instruction-set.md
```diff
@@ -1796,33 +1796,35 @@ Panic if:
...
+| Operation   | `code = match imm { 0 => contract_code(mem[$rA,32]), 1 => blob_payload(mem[$rA,32]), 2 => mem[$ra, ..] }; MEM[$ssp, $rC] = code[$rB, $rC];`       |
 | Syntax      | `ldc $rA, $rB, $rC, imm`                                                                                                                          |
 | Encoding    | `0x00 rA rB rC imm`                                                                                                                               |
-| Notes       | If `$rC` is greater than the code size, zero bytes are filled in.                                                                                 |
+| Notes       | If `$rC` is greater than the code size, zero bytes are filled in. Length final is always padded to word boundary.                                 |
```
Voxelot, line 1807:
```suggestion
| Notes       | If `$rC` is greater than the code size, zero bytes are filled in. Final length is always padded to word boundary.                                 |
```

src/fuel-vm/instruction-set.md
Dentosal: e4ee95b01c286d58230c0c89e621ffe685dc1165

xgreenx (approved):

Issue #456, opened 2023-02-18
LDC support in predicate feature


Marvin8122:
# LDC support

Is it actually practical or sensible to support LDC instraction? How does losing stateless verification affect us? Is it even worth having predicates anymore, if they cannot be verified statelessly?

If we decide to allow LDC, then we could likely support CCP, CROO, CSIZ as well. Maybe storage reads could be supported as well?

_Originally posted by @Dentosal in https://github.com/FuelLabs/fuel-specs/issues/440#issuecomment-1378497483_

SilentCicero:
Related: https://github.com/FuelLabs/fuel-specs/pull/492

PR #612, opened 2024-10-06
LDC support in predicates

Dentosal:
Closes https://github.com/FuelLabs/fuel-specs/issues/456

Requires #611
VM PR https://github.com/FuelLabs/fuel-vm/pull/848

### Before requesting review

### DIFF
--- a/src/fuel-vm/index.md
+++ b/src/fuel-vm/index.md
@@ -82,7 +82,7 @@ To initialize the VM, the following is pushed on the stack sequentially:
 1. Transaction hash (`byte[32]`, word-aligned), computed as defined [here](../identifiers/transaction-id.md).
 1. Base asset ID (`byte[32]`, word-aligned)
 1. [`MAX_INPUTS`](../tx-format/consensus_parameters.md) pairs of `(asset_id: byte[32], balance: uint64)`, of:
-    1. For [predicate estimation](#predicate-estimation) and [predicate verification](#predicate-verification), zeroes.
+    1. For [predicate estimation and verification](#predicate-estimation-and-verification), zeroes.
     1. For [script execution](#script-execution), the free balance for each asset ID seen in the transaction's inputs, ordered in ascending order. If there are fewer than `MAX_INPUTS` asset IDs, the pair has a value of zero.
 1. Transaction length, in bytes (`uint64`, word-aligned).
 1. The [transaction, serialized](../tx-format/index.md).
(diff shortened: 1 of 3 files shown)


src/fuel-vm/index.md
```diff
@@ -157,6 +142,10 @@ For each instruction, its gas cost `gc` is first computed. If `gc > $cgas`, dedu
 
 After the script has been executed, `tx.receiptsRoot` is updated to contain the Merkle root of the receipts, [as described in the `TransactionScript` spec](../tx-format/transaction.md#`TransactionScript`).
 
+### Call
+
+Call context is entered via [`CALL` instruction](./instruction-set.md#call-call-contract). It's also called _internal context_, or _contract context_. Call context is used to access state of a contract.
+
 ## Call Frames
```
xgreenx, line 149: Maybe it is better to do like this?

```suggestion
## Call

Call context is entered via [`CALL` instruction](./instruction-set.md#call-call-contract). It's also called _internal context_, or _contract context_. Call context is used to access state of a contract.

### Call Frames
```

xgreenx (commented):
LGTM=) @Voxelot What do you think?

src/fuel-vm/index.md
```diff
@@ -95,54 +95,39 @@ Then the following registers are initialized (without explicit initialization, a
...
 
 1. `$pc` and `$is` are set to the start of the input's `predicate` field.
-1. `$ggas` and `$cgas` are set to `predicateGasUsed`.
+1. `$ggas` and `$cgas` are set to `MAX_GAS_PER_PREDICATE` for estimation, and `predicateGasUsed` for verification.
 
-Predicate verification will fail if gas is exhausted during execution.
+Predicate execution will fail if gas is exhausted during execution.
 
-During predicate mode, hitting any [contract instruction](./instruction-set.md#contract-instructions) causes predicate verification to halt, returning Boolean `false`.
+During predicate mode, hitting any [contract instruction](./instruction-set.md#contract-instructions) (except `ldc` with non-contract target) causes predicate verification to halt, returning Boolean `false`.
```
xgreenx, line 121: Hmm, it looks like before we should allow usage of blob instructions in the VM as well(because they are not contract instruction)=D So now it works as expected

src/fuel-vm/index.md
Dentosal: No. `Call` is a subheader of `Contexts`.

Voxelot (approved):

PR #613, opened 2024-10-17
Remove erroneous "Notes" from `SUB` and `SUBI` instructions

ironcev:
### Description

This PR fixes the documentation of the `SUB` and `SUBI` instructions and makes it follow the same pattern as the `ADD` and `ADDI` instructions.

The documentation was wrongly using the term "overflow" in the "Notes" of those two instructions, instead of "underflow" (as used in the longer explanation below the instruction).

The PR simply removes the erroneous description from the "Notes" and leaves them empty, as in `ADD` and `ADDI`.

### Before requesting review

### DIFF
--- a/src/fuel-vm/instruction-set.md
+++ b/src/fuel-vm/instruction-set.md
@@ -669,7 +669,7 @@ Panic if:
 | Operation   | ```$rA = $rB - $rC;```                           |
 | Syntax      | `sub $rA, $rB, $rC`                              |
 | Encoding    | `0x00 rA rB rC -`                                |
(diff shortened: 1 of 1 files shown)


Dentosal:
Technically "integer underflow" is still an overflow, i.e. operation resulting a value outside the target domain. That word is widely used to describe this situation, for example in the [llvm docs](https://llvm.org/docs/LangRef.html#llvm-ssub-with-overflow-intrinsics) The note is not incorrect. That said, if it's unclear to you it could be improved.

Dentosal (approved):
Parity with add/addi is a good thing.

Issue #617, opened 2024-11-17
Add "owner" to transaction policies


AurelienFT:
Background:
Currently, each transaction can have many inputs with different owners. We currently have no way to have a "owner" and multiple  implicit rules are used in the ecosystem. 

Proposal:
To give information about a potential "owner" we propose to add a new transaction policy. This new policy would specify an index that will indicate which input of the transaction is considered as the "owner" of the transaction . This feature would enable developers to identify an initiator of the transaction.

Implementation Details:
Add a new optional field owner to the transaction structure.
Update the transaction validation logic to check that the index in the owner field exists in the transaction.

Example:
A transaction with the following policy:

owner: 0x657ab45a6eb98a4893a99fd104347179151e8b3828fd8f2a108cc09770d1ebae

This transaction will provide "0x657ab45a6eb98a4893a99fd104347179151e8b3828fd8f2a108cc09770d1ebae" as the owner and will be accessible in the code executed in this transaction

PR #618, opened 2024-11-19
Add owner policy

AurelienFT:
https://github.com/FuelLabs/fuel-specs/issues/617

VM PR : https://github.com/FuelLabs/fuel-vm/pull/872

### Before requesting review

- [ ] Someone else?


src/tx-format/policy.md
```diff
@@ -63,3 +64,13 @@ Transaction is invalid if:
...
+
+## `Owner`
+
+| name    | type     | description                                              |
+|---------|----------|----------------------------------------------------------|
+| `owner` | `uint64` | Index of the transaction input that we nominate as owner |
+
+Transaction is invalid if:
+
+- `owner >= inputs.len()`
```
Voxelot, line 76: It should also be invalid if the input type doesn't have an owner field.

Ie messages or contract inputs.

If we wanted both coins and messages supported then it should be owner in the case of coin inputs, and receiver in the case or message inputs.

AurelienFT: Mhh okey who can take a decision on what we want to be valid ? Given my little experience I'm ok with both.

Voxelot: It should probably be both messages and coins, given that `coinsToSpend` can select either a coin or message for the base asset.

Voxelot: @xgreenx could you clarify what you mean by this? It's not obvious in the context of design specs.

xgreenx: I meant that the policy is invalid if the index from the owner's policy doesn't point to the `Message` or `Coin` input.

AurelienFT: > I meant that the policy is invalid if the index from the owner's policy doesn't point to the `Message` or `Coin` input.

This is something handle on the sway level right ?

AurelienFT: Updated spec and code in VM PR.

src/tx-format/policy.md
```diff
@@ -63,3 +64,14 @@ Transaction is invalid if:
...
+- `inputs[owner] != InputType::Message && inputs[owner] != InputType::Coin` (owner need to points to an input which has an owner field)
```
Voxelot, line 77:
```suggestion
- `inputs[owner] != InputType::Message && inputs[owner] != InputType::Coin` (must point to an input type that supports ownership)
```

xgreenx (approved):

Issue #615, opened 2024-10-31
Add support for ZK opcodes


xgreenx:
The description of the feature:

https://docs.google.com/document/d/1oQkJ9sJ-YHvcQApaKCYVvXkb9ivUKVoZXn2oSkLoprQ/edit?usp=sharing

## Definition of done
`fuel-vm` is updated and supports new opcodes(in std and non-std context) and defines the gas cost types for it.
`fuel-core` is updated to support new opcodes and has benchmarks to estimate opcodes.

PR #619, opened 2024-11-19
Add ZK opcodes definition

AurelienFT:
https://github.com/FuelLabs/fuel-specs/issues/615

VM PR : https://github.com/FuelLabs/fuel-vm/pull/870

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/src/fuel-vm/instruction-set.md
+++ b/src/fuel-vm/instruction-set.md
@@ -2376,7 +2376,7 @@ Panic if:
 
 |             |                                                     |
 |-------------|-----------------------------------------------------|
-| Description | This opcode regroups all arithmetic operations that can be perform on elliptic curve points. `$rB` defines the curve used. `$rC` defines the type of operation to perform. `$rD` points to the start of the bytes of the operation inputs in memory. `$rA` points to the start of the bytes of the result in memory.                                                   |
+| Description | This opcode regroups all arithmetic operations that can be performed on elliptic curve points. `$rB` defines the curve used. `$rC` defines the type of operation to perform. `$rD` points to the start of the bytes of the operation inputs in memory. `$rA` points to the start of the bytes of the result in memory.                                                   |
 | Operation   | ```MEM[$rA, X] = ecop(MEM[$rD, Y]);```              |
 | Syntax      | `ecop $rA, $rB, $rC, $rD`                           |
 | Encoding    | `0x00 rA rB rC rD`                                  |
--- a/src/fuel-vm/instruction-set.md
+++ b/src/fuel-vm/instruction-set.md
@@ -2375,19 +2375,19 @@ Panic if:
 
 ### `EADD`: Elliptic curve point addition
 
-|             |                                                      |
-|-------------|------------------------------------------------------|
-| Description | The addition of two points (first point bytes starts at `$rC` and second point bytes starts at `$rD`) on `$rB` curve. `$rA` points to the start of the bytes of the addition result. The encoding and decoding depends on the curve (`$rB`) chosen (details below).                         |
-| Operation   | ```MEM[$rA, X] = eadd(MEM[$rC, Y], MEM[$rD, Z]);```  |
-| Syntax      | `eadd $rA, $rB, $rC, $rD`                            |
-| Encoding    | `0x00 rA rB rC rD `                                  |
-| Notes       | For now, only `$rB` = 0 is accepted                  |
+|             |                                                     |
+|-------------|-----------------------------------------------------|
+| Description | The addition of two points (first point bytes starts at `$rC` and second point bytes starts at `$rD`) on `$rB` curve. `$rA` points to the start of the bytes of the addition result. The encoding and decoding depends on the curve (`$rB`) chosen (details below).                        |
+| Operation   | ```MEM[$rA, X] = eadd(MEM[$rC, Y], MEM[$rD, Z]);``` |
+| Syntax      | `eadd $rA, $rB, $rC, $rD`                           |
+| Encoding    | `0x00 rA rB rC rD`                                  |
+| Notes       | For now, only `$rB` = 0 is accepted                 |
 
-#### Curve ID `$rB` possible values : 
+#### Curve ID `$rB` possible values
 
 - `0`: `alt_bn128` elliptic curve.
 
-#### Encoding of points and results by curve ID : 
+#### Encoding of points and results by curve ID
 
 | Curve ID | `$rA` format | `$rC` format | `$rD` format |
 |----------|--------------|--------------|--------------|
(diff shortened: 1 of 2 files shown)


src/fuel-vm/instruction-set.md
```diff
@@ -2370,6 +2372,57 @@ Panic if:
 - `$rB + $rC` overflows or `> VM_MAX_RAM`
 - The memory range `MEM[$rA, 32]`  does not pass [ownership check](./index.md#ownership)
 
+### `ECOP`: Elliptic curve point operation
+
+|             |                                                     |
+|-------------|-----------------------------------------------------|
+| Description | This opcode regroups all arithmetic operations that can be perform on elliptic curve points. `$rB` defines the curve used. `$rC` defines the type of operation to perform. `$rD` points to the start of the bytes of the operation inputs in memory. `$rA` points to the start of the bytes of the result in memory.                                                   |
+| Operation   | ```MEM[$rA, X] = ecop(MEM[$rD, Y]);```              |
+| Syntax      | `eadd $rA, $rB, $rC, $rD`                           |
```
rymnc, line 2381: should it be ecop here?

```suggestion
| Syntax      | `ecop $rA, $rB, $rC, $rD`                           |
```

src/fuel-vm/instruction-set.md
```diff
@@ -2370,6 +2372,57 @@ Panic if:
...
+- `1`: one point and one scalar multiplication
+
+#### Encoding of points and results by curve ID and operation type
+
+- 1P = one point = (X, Y) = ([32 bytes], [32 bytes])
+- 1S = one scalar = X = [32 bytes]
+
+| `$rB` Curve ID | `$rC` Operation type | `$rA` format         | `$rD` format               |
+|----------------|----------------------|----------------------|----------------------------|
+|    `0`         | `0`                  | `MEM[$rA, 64]` `1P`  | `MEM[$rC, 128]` `1P1P`     |
```
rymnc, line 2401:
```suggestion
|    `0`         | `0`                  | `MEM[$rA, 64]` `1P`  | `MEM[$rC, 128]` `1P1S`     |
```

AurelienFT:
@rymnc thank you for pointing incoherences :)

src/fuel-vm/instruction-set.md
```diff
@@ -2370,6 +2372,58 @@ Panic if:
 - `$rB + $rC` overflows or `> VM_MAX_RAM`
 - The memory range `MEM[$rA, 32]`  does not pass [ownership check](./index.md#ownership)
 
+### `ECOP`: Elliptic curve point operation
+
+|             |                                                     |
+|-------------|-----------------------------------------------------|
+| Description | This opcode regroups all arithmetic operations that can be perform on elliptic curve points. `$rB` defines the curve used. `$rC` defines the type of operation to perform. `$rD` points to the start of the bytes of the operation inputs in memory. `$rA` points to the start of the bytes of the result in memory.                                                   |
```
rymnc, line 2379:
```suggestion
| Description | This opcode regroups all arithmetic operations that can be performed on elliptic curve points. `$rB` defines the curve used. `$rC` defines the type of operation to perform. `$rD` points to the start of the bytes of the operation inputs in memory. `$rA` points to the start of the bytes of the result in memory.                                                   |
```

rymnc (approved):
LGTM, small nit

rymnc (approved):

src/fuel-vm/instruction-set.md
```diff
@@ -2370,6 +2372,58 @@ Panic if:
...
+| `$rB` Curve ID | `$rC` Operation type | `$rA` format         | `$rD` format               |
+|----------------|----------------------|----------------------|----------------------------|
+|    `0`         | `0`                  | `MEM[$rA, 64]` `1P`  | `MEM[$rC, 128]` `1P1P`     |
+|    `0`         | `1`                  | `MEM[$rA, 64]` `1P`  | `MEM[$rC, 96]` `1P1S`      |
+
+### `EPAR`: Elliptic curve point pairing check
+
+|             |                                                     |
+|-------------|-----------------------------------------------------|
+| Description | Perform a specific pairing type within a specific curve both identified by `$rB`. `$rC` defines the number of batch of groups. `$rD` define where the bytes of the groups of points start. `$rA` contains either `0` or `1` as the result of the pairing. |
```
Dentosal, line 2408:
```suggestion
| Description | Check if `$rC` groups of points at `$rD` all form valid pairings in (curve, pairing type) identified by `$rB`. Set `$rA` to the result of the pairing, either `0` or `1`. |
```

src/fuel-vm/instruction-set.md
```diff
@@ -2370,6 +2372,58 @@ Panic if:
 - `$rB + $rC` overflows or `> VM_MAX_RAM`
 - The memory range `MEM[$rA, 32]`  does not pass [ownership check](./index.md#ownership)
 
+### `ECOP`: Elliptic curve point operation
+
+|             |                                                     |
+|-------------|-----------------------------------------------------|
+| Description | This opcode regroups all arithmetic operations that can be performed on elliptic curve points. `$rB` defines the curve used. `$rC` defines the type of operation to perform. `$rD` points to the start of the bytes of the operation inputs in memory. `$rA` points to the start of the bytes of the result in memory.                                                   |
```
Dentosal, line 2379:
```suggestion
| Description | Perform arithmetic operation `$rC` on points of the elliptic curve `$rB`. Arguments are read from memory at `$rD`, and is result written to the memory at `$rA`, as per the table below.                                                   |
```

Dentosal (changes requested):
Panic conditions must be described as well. The operations panic at least in cases where mode-selection arguments are not from the list of available modes. Also outputs to memory require that memory to be writeable. Likely there are also cases where the argument values etc. can also be incorrect, in which case the operation should fail. Consider allowing the use of `$err` in those cases.

src/fuel-vm/instruction-set.md
Dentosal: Not sure if my wording is exact enough, but I felt like yours was unnecessarily verbose.

AurelienFT:
@Dentosal I have added panic cases. However, I don't make use of `$err` because I don't really understand in which cases it's better to use that than just make the call fail. 
Following the docs here and a comment on the VM, it seems to be used only on ALU and for me on these opcodes there is no use case for that.

AurelienFT:
I don't think it's expected and I don't see any logic of modifying the input to retry directly designed in code. Also if the point isn't corrected created in the VM implementation then the addition can't be tried and the code will be more difficult and I think users whould except a panic and this will help them debugging with an explicit reason.

rymnc (dismissed):

Dentosal (approved):

